PR #62, opened 2015-03-30
buffers.pl v5.1: merged buffers: always indent

edsantiago:
...not just in left/right layouts: the indentation is visually
helpful in top/bottom too. Removing the conditional even makes
for cleaner code.

### DIFF
--- a/perl/buffers.pl
+++ b/perl/buffers.pl
@@ -20,6 +20,8 @@
 #
 # History:
 #
+# 2015-03-29, Ed Santiago <ed@edsantiago.com>
+#     v5.1: merged buffers: always indent, except when filling is horizontal
 # 2014-12-12
 #     v5.0: fix cropping non-latin buffer names
 # 2014-08-29, Patrick Steinhardt <ps@pks.im>:
@@ -164,7 +166,7 @@ use strict;
 use Encode qw( decode encode );
 # -----------------------------[ internal ]-------------------------------------
 my $SCRIPT_NAME = "buffers";
-my $SCRIPT_VERSION = "5.0";
+my $SCRIPT_VERSION = "5.1";
 
 my $BUFFERS_CONFIG_FILE_NAME = "buffers";
 my $buffers_config_file;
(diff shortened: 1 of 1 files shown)


perl/buffers.pl
```diff
@@ -20,6 +20,8 @@
 #
 # History:
 #
+# 2015-03-29, Ed Santiago <ed@edsantiago.com>
+#     v5.1: merged buffers: always indent, even in top/bottom layouts
```
flashcode: You forgot to update the line with version: `my $SCRIPT_VERSION = "5.0";`

perl/buffers.pl
```diff
@@ -1298,8 +1300,7 @@ sub build_buffers
             }
             else
             {
-                my $indent = "";
-                $indent = ((" " x length($buffer->{"number"}))." ") if (($position eq "left") || ($position eq "right"));
+                my $indent = (" " x length($buffer->{"number"}))." ";
```
flashcode: In the commit 0149cb9be5f1e2b50de9d54694f4a843a8c378b8, I disabled the indent on top/bottom position, and I think we should keep that with a small change. The indent should be enabled on top/bottom ONLY if the filling is not "horizontal". In other words, the indent should be enabled everywhere, except if you have position top/bottom with horizontal filling.

perl/buffers.pl
edsantiago: Whoops. Thanks for catching that.

perl/buffers.pl
edsantiago: Ahhhh... yes, I see. Thanks again. I'll have to play with that setting and see how to adjust for it.

edsantiago:
I have a fix. Would you prefer a new commit (2 total), or (what I tend to consider cleaner) git commit --amend && git push --force?

flashcode:
You can rebase your commits into a single one, then git push --force.

perl/buffers.pl
```diff
@@ -1298,8 +1300,17 @@ sub build_buffers
...
-                my $indent = "";
-                $indent = ((" " x length($buffer->{"number"}))." ") if (($position eq "left") || ($position eq "right"));
+                # Indentation aligns channels in a visually appealing way
+                # when viewing list top-to-bottom...
+                my $indent = (" " x length($buffer->{"number"}))." ";
+                # ...except when list is top/bottom and channels left-to-right.
+                my $option_pos = weechat::config_string( weechat::config_get( "weechat.bar.buffers.position" ) );
+                if (($option_pos eq 'top') || ($option_pos eq 'bottom')) {
+                    my $option_filling = weechat::config_string( weechat::config_get( "weechat.bar.buffers.filling_top_bottom" ) );
+                    if ($option_filling eq 'columns_horizontal') {
```
flashcode: I would compare to "horizontal", no?
For me any filling with columns should have indent, only "horizontal" should not have it.

edsantiago:
Wow, this has become complicated! I never realized there were so many layout options. After much playing I believe indent should be disabled in _all_ horizontal cases (horizontal and columns_horizontal). In columns_horizontal I believe indenting gives the false impression that the channels are associated vertically.

In case anyone would like to confirm, I wrote a quick tool to help me cycle through possibilities:
https://gist.github.com/edsantiago/f45ef3743cfe8c5d4931

flashcode:
Merged, thanks.

PR #64, opened 2015-04-14
Improve buffer finding mechanics

torhve:


flashcode:
It was merged on Apr 14th.

PR #76, opened 2015-06-07
Strip Colour Codes from Title

t3chguy:
Remove junk from title which doesn't belong there - Colours.

### DIFF
--- a/python/title.py
+++ b/python/title.py
@@ -22,6 +22,8 @@
 # (this script requires WeeChat 0.3.0 or newer)
 #
 # History:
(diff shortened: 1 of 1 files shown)


flashcode:
Hi,
Your fix is on the title var, which is in fact the name or short name of the buffer (the channel name, or the name displayed).
Are you sure there are color codes inside the channel name?

t3chguy:
It can be just on the shortname, but I felt this looked more Pythonic
If you prefer it to just be on the shortname then say so

flashcode:
Merged (I reworded the commit message).

PR #78, opened 2015-06-29
Use last index to handle all buffer nesting levels

MatthewCox:
This change makes the script work without errors no matter the nesting level of the buffer you are highlighted in.

My actual use-case for this is with the weetweet.py script, which adds a server-level 'twitter' buffer.
Any time I get @mentioned on Twitter currently I get an error in the weechat buffer due to the invalid indexing of $chan.


sn0opy:
Haven't thought about something like that, when I created the script. Feel free to merge.

flashcode:
Merged.

PR #87, opened 2015-08-01
fish.py: Added cbc support for channels and users (does not work for key exchanges yet).

d347hbr1ng3r:


flashcode:
Hi,
Did you send the pull request on the upstream repository? https://github.com/freshprince/weechat-fish
If accepted by the author of script, then a pull request can be opened here (with a single commit using the appropriate format and with updated version inside script, see the file Contributing.asciidoc in this repository for more info).

PR #88, opened 2015-08-03
otr.py 1.7.0: Version 1.7.0 enhancements and bug fixes.

mmb:
Allow setting policies on a per server and per account basis. Peer policies
override account policies which override server policies.

Convert data loaded from trust files to unicode.

Do not OTR whitespace tag nicks starting with *.

Nicks to skip OTR tagging for are now a user-configurable regex. Set the
general.no_send_tag_regex option.

Add otr.general.default.key option to share a key across accounts. Set
this to nick@server of the account with the key you want to use as the
default key for new accounts.

Do not OTR whitespace tag CTCP messages.

### DIFF
--- a/python/otr.py
+++ b/python/otr.py
@@ -35,6 +35,7 @@ import platform
 import re
 import traceback
 import shlex
+import shutil
 import sys
 
 class PythonVersion2(object):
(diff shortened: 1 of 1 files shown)


flashcode:
Merged.

PR #89, opened 2015-08-06
[oldswarner] FIX urls are not patterns!

boyska:
Without this patch, url could get searched as if they were pattern.
The "plain=true" parameter will fix this.


torhve:
+1
Thanks!

PR #103, opened 2015-09-28
buffer_autoset.py 0.9: instantly apply properties

sim642:
Previously buffer_autoset.py required buffers to be reopened for the set options to apply or the user had to also run an extra `/buffer set` command. This change makes the script listen to the script's config options and applies them immediately, even when modifying the settings directly not via the script's command. It does not do anything when the option gets removed because there's no way to retrieve the supposed value.

*Tested by me and strayPuppy from #weechat.*

### DIFF
--- a/python/buffer_autoset.py
+++ b/python/buffer_autoset.py
@@ -22,6 +22,8 @@
 #
 # History:
 #
+# 2015-09-28, Simmo Saan <simmo.saan@gmail.com>:
+#     version 0.9: instantly apply properties
 # 2015-07-12, Sébastien Helleu <flashcode@flashtux.org>:
 #     version 0.8: add option buffer_autoset.look.timer to add a small timer
 #                  before setting buffer properties
@@ -44,7 +46,7 @@
 
 SCRIPT_NAME = "buffer_autoset"
 SCRIPT_AUTHOR = "Sébastien Helleu <flashcode@flashtux.org>"
-SCRIPT_VERSION = "0.8"
+SCRIPT_VERSION = "0.9"
 SCRIPT_LICENSE = "GPL3"
 SCRIPT_DESC = "Auto-set buffer properties when a buffer is opened"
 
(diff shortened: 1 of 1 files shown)


de-vri-es:
Interesting, I kinda wanted something similar. However, applying it instantly might be a bit too eager? Alternatively, a subcommand could be added to apply all rules to the current buffers. Or you could make it configurable whether or not this is done automatically. Or even better, have a subcommand and an option to do it automatically. Just a few thoughts.

sim642:
I thought about not making it instant always (like an option or subcommand) but what would be a scenario where instantly applying the option would not be desired? Or would have bad effects?

de-vri-es:
Well, you can test the rules with new buffers first, then apply to all matching buffers later.

highflyer77:
This is a nice addition to the script!  Now there is no need to do 2 commands or reload the buffer.

Like sim642 said, there is no scenario where instantly applying would not be disired.  That's the purpose of autoset_buffer in the first place. Also, you can still use /buffer set "foo" if you don't wish to save it. 
(nice side affect is: you don't have to wait till you reload the buffer to see if you entered the autosetbuffer command correctly).

@sim642  Nice Job

de-vri-es:
I agree, nice addition. I do think that testing rules on new buffers first has value though, especially since you can't auto unset after removing a rule.

sim642:
@de-vri-es If you previously tested rules on some other buffer, you still needed to reopen the buffer to get rid of the changes on that test buffer. So how does this break that workflow?

de-vri-es:
The test buffer could be some new buffer you are going to delete anyway. But if the rule is applied to all matching buffers immediately you have to close and reopen all affected buffers which is much more work than restoring one buffer.

Don't get me wrong, I really like the idea, but I would be hesitant with doing things automatically without having an option.

sim642:
If you have a special test buffer for it anyway, why would you write rules to apply to some other buffers other than your test one?

sim642:
Either way, I just changed the PR and added an option which allows disabling of this instant behavior. It only affects changes made when it's `on`, so changing some autosets and turning the option from `off` to `on` won't cause them all to apply. I'd say this allows for fine tuning when to use instant application and when not.

de-vri-es:
Cool, thanks

de-vri-es:
Oh, one small thing, I noticed the script is tagged py3-ok, but string formatting with the % operator is not supported in python 3 anymore. It should probably be replaced with `string.format()`.

Again, thanks for the work, nice feature.

sim642:
I just used % operator syntax because it's used all over the rest of the script's code too. It probably could be fixed (either the tag or the code) but I'd say that's out of the bounds of this PR.

de-vri-es:
Ah, good point, better done separately.

flashcode:
The % operator is still supported in python 3.

de-vri-es:
Oh I see, it's just deprecated, not actually removed.

PR #106, opened 2015-10-11
autojoin_on_invite.py 0.6: allow joining channels with keys in autojoin (closes #99)

### LINKED ISSUE #99: autojoin_on_invite.py, add feature to autojoin +keyed channels
please make weechat use the irc.server.*.autojoin key when invited to a +keyed channel. currently weechat will not use it and therefor not autojoin a channel that has key protection.

sim642:
This change makes the script use channel keys from server's `autojoin` option when joining a channel (disable-able via an option).

### DIFF
--- a/python/autojoin_on_invite.py
+++ b/python/autojoin_on_invite.py
@@ -20,6 +20,8 @@
 # (this script requires WeeChat 0.3.0 or newer)
 #
 # History:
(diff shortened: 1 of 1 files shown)


motharussia:
Change confirmed. +k channels are now autojoined using saved channel key.

Issue #114, opened 2015-11-12
lnotify 0.3.0 misbehaves and notifies user's messages itself


ersuniljoshi:
Xubuntu 15.10 
Linux shupkarn 4.2.0-18-generic #22-Ubuntu SMP Fri Nov 6 18:25:50 UTC 2015 x86_64 x86_64 x86_64 GNU/Linux
WeeChat 1.4-dev (git: v1.3-103-g99ebcca) [compiled on Nov 12 2015 00:14:04]

I have upgraded lnotify from 0.2.0 to 0.3.0 and started facing this issue
I am using bitlbee (3.4.1) + weechat
bitlbee-facebook was installed to interact with facebook.

As soon as i typed messages to my friends, they itself got highlighted to me itself.
I have debugged the issue on #weechat but finally after reverting the script to 0.2.0 version, above mentioned problem got resolved. Let me know if anything else is needed from my end on this.

PR #115, opened 2015-11-12
Fix issue #114 with lnotify.py

nullsym:
As noted by ersuniljoshi on https://github.com/weechat/scripts/issues/114 the newest version of lnotify.py notified users of messages that they sent. The user presumably only wants to receive notifications from messages that they receive, not on the messages that they sent. This commit fixes such behaviour.


flashcode:
Merged, thanks.

PR #126, opened 2015-12-23
whatsapp.py version 0.1

jspricke:
- Single user chat working.
- /wchat -> /query


flashcode:
Hi,
For new scripts, please don't make a pull request (as written in Contributing.asciidoc), but use the form at: https://weechat.org/scripts/add/
Thanks.

jspricke:
Done. I though it would have been nice to track the history of the script as well, as I copied it from jabber.py and it still shares a lot of code with it. So it may be easier to keep both of them in sync.

PR #129, opened 2015-12-27
weemoticons.py 0.3: fix regex matching for urls and new smileys

Ratler:


flashcode:
Hi,
Why not using version 0.2 for the update?
And if possible, please keep style as-is on existing code (like quotes), so that the diff contains only your changes.
Thanks.

Ratler:
@flashcode sorry for the delay, I had an unofficial 0.2 for quite some time so I thought it might have been a good idea to go with 0.3. I can of course change it to 0.2 if you feel it's necessary.

flashcode:
I already merged #128, this one is the same?

Ratler:
Ah yes, the only difference was the commit message.

PR #143, opened 2016-02-20
histsearch.py 0.3: add option to only display selected command

sim642:
Usually input bar's size is set to `1` and with many long inputs this script's output is impossible to see as it just doesn't fit the input bar. The way this PR solves the issue is by adding an additional option allowing the display of only the currently selected historical command instead of the whole list, fitting into the input bar much more likely. It also is more like bash's behavior.

*Feature requested and PR confimed working by TreyHarris in #weechat.*

### DIFF
--- a/python/histsearch.py
+++ b/python/histsearch.py
@@ -24,6 +24,8 @@
 # (this script requires WeeChat 0.3.0 or newer)
 #
 # History:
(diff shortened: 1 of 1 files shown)


sim642:
@torhve Offically the contribution guide requires me to run these changes by you.

PR #144, opened 2016-02-29
emoji.lua 1: New script.

torhve:
New script to allow emoji transformations in text. Like :heart:
Also adds tab complete :hea<tab> => ❤


flashcode:
For a new script, please use the form at: https://weechat.org/scripts/add/ (this initializes an entry in the scripts database, with the pending script).

PR #152, opened 2016-05-01
buffers.pl 5.4: added option "detach_buffer_immediately_level"

mumixam:
This new option controls what messages cause a buffer to be attached that was detached when option "detach_buffer_immediately" is enabled based on message notify level

### DIFF
--- a/perl/buffers.pl
+++ b/perl/buffers.pl
@@ -20,6 +20,8 @@
 #
 # History:
 #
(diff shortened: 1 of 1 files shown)


mumixam:
@rgrinberg you disappeared from #weechat when i started working on this. Have you tried my changes? Does it work as excepted?

rgrinberg:
@mumixam sorry for disappearing. Trying this out now and it works like a charm

Thank you!

PR #155, opened 2016-05-12
keepnick.py 1.3: monitor quits and nick changes

pix0r:
Also fixed what appeared to be a bug where when the user has multiple nicks for a server, redirect_isonhandler would attempt to regain each nick in the list.

PR also submitted to https://github.com/weechatter/weechat-scripts/pull/19


flashcode:
Closing, implemented via 7ea5f610104e02175777116704594f9de73b6925.

PR #161, opened 2016-06-21
grep 0.7.7: fix get_home function

mickael9:
With the default settings, the `get_home` function returns an invalid path such as:
`/home/user//home/user/.weechat/logs`

This is caused by calling `abspath` on a path that starts with a `%h` placeholder which turns the path into an absolute path from the working directory (`/home/user/%h/logs`).

This uses the new `string_eval_path_home` function which does this correctly.

### DIFF
--- a/python/grep.py
+++ b/python/grep.py
@@ -66,6 +66,9 @@
 #
 #   History:
 #
+#   2016-06-23, mickael9
+#   version 0.7.7: fix get_home function
+#
 #   2015-11-26
 #   version 0.7.6: fix a typo
 #
@@ -203,7 +206,7 @@ except ImportError:
 
 SCRIPT_NAME    = "grep"
 SCRIPT_AUTHOR  = "Elián Hanisch <lambdae2@gmail.com>"
-SCRIPT_VERSION = "0.7.6"
+SCRIPT_VERSION = "0.7.7"
 SCRIPT_LICENSE = "GPL3"
 SCRIPT_DESC    = "Search in buffers and logs"
 SCRIPT_COMMAND = "grep"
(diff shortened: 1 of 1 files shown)


sim642:
Most of the logic in the `get_home` function could simply be performed by a single call to [`weechat.string_eval_path_home`](https://weechat.org/files/doc/devel/weechat_plugin_api.en.html#_string_eval_path_home). There's no need to duplicate (currently incorrectly) functionality which already exists in the API.

mickael9:
@sim642  Good point, I updated the PR to use that

m4v:
string_eval_path_home is for weechat 1.3, while grep.py states in its header that supports weechat 0.3.* and I don't really like the idea of killing older version support just because of a string manipulation.

I can't test but I think inverting the other of functions should fix this issue, replace '%h' first then make it go through abspath()

mickael9:
This was my first proposition
```diff
 def get_home():
     home = weechat.config_string(weechat.config_get('logger.file.path'))
+    home = home.replace('%h', weechat.info_get('weechat_dir', ''))
     home = path.abspath(path.expanduser(home))
-    return home.replace('%h', weechat.info_get('weechat_dir', ''))
+    return home
```

Do you want me to update this PR to this version again?

m4v:
>Do you want me to update this PR to this version again?

yes.

mickael9:
Here you go!

python/grep.py
```diff
@@ -417,8 +420,7 @@ def get_config_log_filter():
 
 def get_home():
     home = weechat.config_string(weechat.config_get('logger.file.path'))
-    home = path.abspath(path.expanduser(home))
-    return home.replace('%h', weechat.info_get('weechat_dir', ''))
+    return weechat.string_eval_path_home(home, '', '', '')
```
mickael9: Why does this show the previous changeset? The fix asked by @m4v doesn't use `weechat.string_eval_path_home` anymore.

flashcode: I started a review on your old code, now the string_eval_path_home has gone with your latest changes, so you can just ignore my comment.

PR #140, opened 2016-01-28
go.py 2.2: Add option fuzzy_search

ylambda:
Adds an option for fuzzy searching buffers.

image

PR #165, opened 2016-07-14
go.py 2.2: add multiple consecutive term matching + highlighting

gilbertw1:
This pull request adds the ability to specify multiple matching terms to the go.py script in order to easily uniquely identify a particular channel.

Having added several slack servers to weechat, I am noticing (especially with 'general' channels) that I often have a lot of overlap between channel names. This fact makes it difficult to jump directly to the channel I want using go.py. 

My solution (taken from the [z directory jumper](https://github.com/rupa/z/wiki#spend-a-lot-of-time-cd-ing-around-a-complex-directory-tree)) is to allow multiple terms to be used to make it easier to uniquely identify a channel. This works by matching each term one after another while ensuring that each term occurs in the matching channel after the previous one. The easiest way to illustrate this is through the following gif:

weechat-go

I will mention that per the contributing guidelines, I did attempt to email the author of this script, but did not get a response. Also, seeing there is no separate repository hosting this plugin, I am submitting this PR directly. 

Thanks for all your hard work!

### DIFF
--- a/python/go.py
+++ b/python/go.py
@@ -21,6 +21,9 @@
 #
 # History:
 #
+# 2016-07-13, Bryan Gilbert <gilbertw1@gmail.com>
+#     version 2.2: add multiple sequential term matching to uniquely identify
+#                  channels
 # 2015-11-12, nils_2 <weechatter@arcor.de>
 #     version 2.1: fix problem with buffer short_name "weechat", using option
 #                  "use_core_instead_weechat", see:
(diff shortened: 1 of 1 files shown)


flashcode:
Hi,

Does the current version 2.2 (done via PR #140) solve your problem?
If yes, we can close this PR.

gilbertw1:
Yep! Thanks!

PR #169, opened 2016-09-20
topicdiff.py 0.3: improve diff output

mickael9:
- Keep the order of chunks
- Move the '-'/'+' symbols into the prefix and color them red/green
- Allow '~' and '-' as chunk separators

Sample output:

image

### DIFF
--- a/python/topicdiff.py
+++ b/python/topicdiff.py
@@ -4,11 +4,13 @@
 
 import re
 
+from itertools import izip_longest
+
 import weechat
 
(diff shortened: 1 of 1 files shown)


mickael9:
> Hi Mickaël,
>
> These changes seem reasonable but I don't have any commit bits on
> weechat (or even a github account). If you need my informal approval,
> you have it, but maybe it makes more sense for someone more active to
> assume maintainership. At least, I don't want to stand in the way of
> further improvements.
>
> Regards,
>
> Daf

PR #182, opened 2016-12-24
gnome_screensaver_away.py 0.2.0: don't trample manual away

grdryn:
- Prevent changing manually-set away status
- Allow configuration of message and poll time
- Tolerate gnome-shell crashes


grdryn:
Modified to tolerate gnome-shell crashes

Issue #188, opened 2017-01-30
urlserver.py: incorrectly includes non-percent-encoded characters in URIs


MacGyverNL:
Per [RFC3986, section 2](https://tools.ietf.org/html/rfc3986#section-2), any character not in either the reserved set or the unreserved set in a URI *must* be percent-encoded. Characters for which is the case are e.g. the space (%20), but also " (%22), < (%3C), > (%3E), { (%7B), and } (%7D). As it stands, urlserver.py seems to limit the end-match character class to the characters ], ), >, and the whitespace class \s. This is both incomplete and wrong: incomplete because it ignores such characters as " and <, and wrong because ) is a valid sub-delim for any URI scheme and thus cuts off many valid URLs.

This leads to annoying situations where e.g. shortened embedded tweet URIs from twitter's URI title scheme include the " in the redirect (twitter adds "s around the entire tweet in the title text of the page, so any tweet ending in a URI gets a wrong short link). Since title-pasting bots are rather common, so does this problem seem to be.

Another annoying problem is that valid wikipedia links are often cut before the end, since a lot of them include parentheses (), e.g. any page with a disambiguation (<https://en.wikipedia.org/wiki/Git_(disambiguation)>, and <https://en.wikipedia.org/wiki/Git_(slang)>).

As far as I can tell, [RFC3987](https://tools.ietf.org/html/rfc3987) (internationalized resource identifiers) does not change this situation.

My first thought to fix this is to change the matching regex to a positive rather than a negative match, and simply positive-match all the characters from the unreserved and reserved sets. However, this will probably not work for internationalized URLs. Rather, then, a negative-match on *all* the ASCII-characters not in the reserved and unreserved sets should work.

This may lead to the undesirable characteristic where people delimiting URLs with parentheses find their closing parenthesis matched by the regular expression. I'll state once again that they're doing it wrong, but since it is such a natural thing to do, maybe another match should be performed and if there exists a matching opening parenthesis before the URI scheme specifier for a closing parenthesis at the end of the URI, the closing parenthesis should be dropped from the match. Another option, that the URXVT terminal matcher.pl seems to prefer, is to include the closing parenthesis as part of the URI if and only if there is a matching opening parenthesis present anywhere in the URI. This could be limited to the part after the domain specifier, and could be part of a single negative-matching regex.

MacGyverNL:
[...]

PR #189, opened 2017-02-02
urlserver.py 2.4: better matching

MacGyverNL:
The intention of this branch is to fix the issues identified in #188. It does this in a twofold way:

1. It intends to match _anything_ that looks like a valid URI (with a few exceptions; e.g. it requires the :// specifier, where RFC3986 has three other forms of URIs; these don't seem to be widely used) using a regular expression based on the ABNF grammar specified in [RFC3987, section 2](https://tools.ietf.org/html/rfc3987#section-2). I'm highly confident this grammar specifies a regular language[1], so using a regular expression that actually matches the language, rather than a superset thereof, should provide complete and correct matching; no more cutting off URIs too early.
However, there's a few cases in which I deviate from the grammar, such as with "[" and "]". Rationale in the commit log.

2. It uses non-regex heuristics in the matching function to fix the common mistake of delimiting URIs with characters valid in a URI. Currently handled are `.` `,` `)` `]` `'`.
(Maybe this should be in a different function altogether which gets passed the MatchObject.)
Currently this does not recurse, so it doesn't handle it if somebody uses _two_ such characters in sequence, e.g. "the full quote is `http://www.example.example/example.html.' -- author". I don't think that's a bad thing but I've seen at least one case of it. Might extend the heuristics for that for certain cases.

All in all, I think this improves the matching significantly and, at least for me, it seems to remove URLs cut off beforehand altogether, and reduces URLs that are overcomplete (which are less of a problem, as long as it doesn't happen too often, since visual inspection usually allows an easy manual fix) to a minimum.

[1] Michael Sperberg-McQueen has a computer-assisted proof of this for RFC3986 at http://cmsmcq.com/mib/?p=306, I don't believe RFC3987 introduces recursion in the grammar so it should hold for that as well.

### DIFF
--- a/python/urlserver.py
+++ b/python/urlserver.py
@@ -44,6 +44,8 @@
 #
 # History:
 #
+# 2018-10-01, Pol Van Aubel <dev@polvanaubel.com>:
+#     v2.4: rework URL matching to positive regex match with heuristics
 # 2018-09-30, Sébastien Helleu <flashcode@flashtux.org>:
 #     v2.3: fix regex in help of option "http_allowed_ips"
 # 2017-07-26, Sébastien Helleu <flashcode@flashtux.org>:
@@ -114,7 +116,7 @@
 
 SCRIPT_NAME = 'urlserver'
 SCRIPT_AUTHOR = 'Sébastien Helleu <flashcode@flashtux.org>'
-SCRIPT_VERSION = '2.3'
+SCRIPT_VERSION = '2.4'
 SCRIPT_LICENSE = 'GPL3'
 SCRIPT_DESC = 'Shorten URLs with own HTTP server'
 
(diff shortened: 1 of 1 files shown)


MacGyverNL:
Fixed a bug with indexing errors in the heuristics when a matched URL starts at the beginning of a message. Considering the lack of comments, I've now squashed everything into a single mergeable commit. The feature branch is still available as [urlserver-compliance-unsquashed](https://github.com/MacGyverNL/weechat-scripts/tree/urlserver-rfccompliance-unsquashed).

anonymous2ch:
Can you apply the same matching to url_olde.py please?

MacGyverNL:
Rebased onto 2.3. I've been running this code without problems for over 1.5 years now.

flashcode:
Thanks, merged!
And sorry for the (very) long delay…

PR #190, opened 2017-02-21
buffers.pl 5.5: fix memory leak in perl 5.23.7-5.24.1, fix truncation…

arza-zara:
… and crop_suffix when truncating to 1-4 characters, fix prefix_empty for inactive buffers, tidy code


flashcode:
Thanks, merged.

PR #192, opened 2017-02-25
go.py 2.3: fix fuzzy search breaking buffer number search display (closes #191)

### LINKED ISSUE #191: go.py 2.2: fuzzy search breaks display when inputting buffer number
Fuzzy search **off** and `alt+g` followed by input `1`: 
screenshot - 24 02 2017 - 19 14 05_2

Fuzzy search **on** and `alt+g` followed by input `1`: 
screenshot - 24 02 2017 - 19 14 23_2

sim642:
An extra fuzzy match check was added to displaying the search results to avoid buffers matched by number from being rendered using fuzzy search display methods which fail in this case due to the number not actually being present in the buffer name.

### DIFF
--- a/python/go.py
+++ b/python/go.py
@@ -21,6 +21,8 @@
 #
 # History:
 #
+# 2017-02-25, Simmo Saan <simmo.saan@gmail.com>
+#     version 2.3: fix fuzzy search breaking buffer number search display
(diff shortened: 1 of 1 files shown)


sim642:
Sorry for the trouble, I must've reformatted it after loading the script for testing, forgetting that Python needs a \ at the end of a line to break it into multiple.

PR #195, opened 2017-03-10
pushover.pl 2.0: fix pushover.net API call

stfnm:
Also dropped support for the other notification services in favor
of the new hsignal "pushover".

This can be used with any /trigger (for example combined with /exec)
or from other weechat scripts. That way it's very easy to use different
different notification services without having to reimplement the entire
"when to notify" logic and without bloating this script.

See /help pushover for trigger examples.


stfnm:
See also https://github.com/stfnm/weechat-scripts/issues/23 for reference.

PR #206, opened 2017-04-08
Update url_olde.py

anonymous2ch:
Make it match URLs containing UTF-8 chars without percent encoding, like in this list: https://mathiasbynens.be/demo/url-regex


flashcode:
Merged (version updated to 0.6).

PR #208, opened 2017-04-12
fish 0.9.2: also put mark on outgoing messages

rr-:
This substitutes the message before it ends up being sent to the buffer.
After the message with the mark is drawn in the buffer, the later `PRIVMSG` hook strips the marker back before encrypting and sending it to the network.


rr-:
Bug: because the script doesn't try to be too smart and doesn't try to parse any commands, the outgoing text sent with `/say` and `/msg` ends up having no marker (when it should).
I'd argue that it's an acceptable tradeoff at this stage since these commands aren't used very often.

PR #213, opened 2017-05-04
shortenurl.py 0.6.1: fix api call to is.gd

jsaalfeld:
I talked in IRC with someone about this Bug. It seems, that is.gd changed the API call, and there was immediately a fix for this.

### DIFF
--- a/python/shortenurl.py
+++ b/python/shortenurl.py
@@ -14,6 +14,8 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 # History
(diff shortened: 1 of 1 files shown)


jsaalfeld:
@flashcode: what info are you waiting for?

flashcode:
@jsaalfeld : I started a review because for me there's a bug with the variable ISGD, my message was:

> I think this will not work when doing ISGD % encode(...) because there's no % in the string: the formatting will fail.

jsaalfeld:
@flashcode I updated the code and squashed it into one commit again. Should be fine now. Thanks for pointing that out!

PR #215, opened 2017-05-07
weetweet.py 1.2.6: Add support for py3 weechat

DarkDefender:
The script now works if weechat is running either python 2 or 3. It previously only worked with python2.
(However, the script still always requires python3 for the secondary process it spawns)

So the script should now be "py3k-ok"

Is there a file that I should add that tag to or is it out of my hands?


flashcode:
Thanks, merged.
And tag "py3k-ok" added in database.

PR #218, opened 2017-05-14
Buffers.pl: Add option buffers.look.hide_inactive

KellerFuchs:
This options allows hiding buffers that do not contain new messages, which is useful when having many buffers open simultaneously.

@arza-zara, would you be willing to review this pull request?

### DIFF
--- a/perl/buffers.pl
+++ b/perl/buffers.pl
@@ -650,6 +650,12 @@ my %default_options_color =
 
 my %default_options_look =
 (
+ "hide_inactive" => [
+     "hide_inactive", "boolean",
+     "Skip inactive buffers when on.",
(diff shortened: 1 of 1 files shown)


arza-zara:
You can `/set buffers.look.detach 1` and they are hidden after one second.
In buflist (WeeChat 1.8) you can use a condition:
```
/set buflist.look.display_conditions ${buffer.hotlist.priority}>0||${buffer.num_displayed}>0 && ${buffer.hidden}==0
```

KellerFuchs:
@arza-zara That indeed works.
At first, I misunderstood (due to some `buflist` vs. `buffers.pl` confusion)

PR #224, opened 2017-06-02
spotify.py 0.8: add now required oauth support

butlerx:
Ive emailed the author xt@bash.no about this change


butlerx:
ive realised the token expire quite quickly so this method isnt very usedfull so im closing this pr

PR #233, opened 2017-08-08
url_hint fix unexpected encoding errors

vmindru:
when url_hinter encounters "weired" encodings in chat messages it will fail with and unexpected encoding message, this will convert "best effort" all text to UTF-8 and try to find an URL.


flashcode:
Merged, thanks.

PR #237, opened 2017-09-09
launcher.pl 0.7: properly fix escaping of "signal_data"

Hello71:
This time I actually tested it.


flashcode:
Thanks, merged (I updated the version as well).

PR #239, opened 2017-09-20
grep.py 0.8

mickael9:
- use weechat 1.5+ api for background processing (old method was unsafe and buggy)
- add timeout_secs setting (was previously hardcoded to 5 mins)

### DIFF
--- a/python/grep.py
+++ b/python/grep.py
@@ -53,6 +53,9 @@
 #     It can be used for force or disable background process, using '0' forces to always grep in
 #     background, while using '' (empty string) will disable it.
 #
+#   * plugins.var.python.grep.timeout_secs:
+#     Timeout (in seconds) for background grepping.
+#
 #   * plugins.var.python.grep.default_tail_head:
 #     Config option for define default number of lines returned when using --head or --tail options.
 #     Can be overriden in the command with --number option.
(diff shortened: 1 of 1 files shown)


python/grep.py
```diff
@@ -66,6 +69,11 @@
 #
 #   History:
 #
+#   2017-09-20, mickael9
+#   version 0.7.9:
+#   * use weechat 1.5+ api for background processing (old method was unsafe and buggy)
```
Zarthus: Does this change the minimum weechat script version possible? Version bump should at least be minor (or perhaps even major, due to bc break) - definitely not patch.

python/grep.py
```diff
@@ -198,7 +206,7 @@
 ###
 
 from os import path
-import sys, getopt, time, os, re, tempfile
+import sys, getopt, time, os, re, cPickle
```
Zarthus: cPickle seems to be a python 2 thing. In python 3 it's called pickle.

Perhaps you should do something like 

```py
pycompat = sys.version_info[0] < 3
if pycompat:
  import cPickle as pickle
else:
  import pickle

# and adapt the code to work in both py2 and py3 where API changes are necessary.
```

Zarthus: Without this compatibility it is a breaking change as I think grep works on python 3. If it doesn't, movement towards 3.x support is still something good to be working towards.

python/grep.py
mickael9: I bumped it to 0.8, but I think users should be discouraged from using weechat <1.9.1 with all the security vulnerabilities

python/grep.py
mickael9: I added a fallback to `pickle` in case importing `cPickle` fails

PR #240, opened 2017-10-01
spotify.py version 0.9: Add support for oauth keys being stored in secure data

butlerx:
I've emailed xt as script owner, waiting for a response

### DIFF
--- a/python/spotify.py
+++ b/python/spotify.py
@@ -1,6 +1,6 @@
 # -*- coding: utf-8 -*-
 #
-#Copyright (c) 2009 by xt <xt@bash.no>
+# Copyright (c) 2009 by xt <xt@bash.no>
 #
 # This program is free software; you can redistribute it and/or modify
 # it under the terms of the GNU General Public License as published by
@@ -24,6 +24,8 @@
 #
 #
 # History:
+# 2017-09-30, butlerx
+#   version 0.9: Add support for oauth keys being stored in secure data
 # 2017-06-02, butlerx
 #   version 0.8: add now required oauth support
 # 2016-01-22, creadak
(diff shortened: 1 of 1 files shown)


python/spotify.py
```diff
@@ -43,48 +45,66 @@
...
-    "buffers"       : 'A comma separated list of buffers the script should check',
-    "emit_notice"   : 'If on, this script will use /notice, if off, it will use /msg to post info',
-    "client_id"     : 'required client id token go to https://developer.spotify.com/my-applications/#!/applications to generate your own',
-    "client_secret" : 'required client secret token go to https://developer.spotify.com/my-applications/#!/applications to generate your own'
+    "buffers":
+    'A comma separated list of buffers the script should check',
+    "emit_notice":
+    'If on, this script will use /notice, if off, it will use /msg to post info',
+    "client_id":
+    'required client id token go to https://developer.spotify.com/my-applications/#!/applications to generate your own',
```
Zarthus: What was your line limit restriction when reformatting the code? If it was 80 this exceeds said limit.

python/spotify.py
```diff
@@ -43,48 +45,66 @@
...
+    "client_secret":
+    'required client secret token go to https://developer.spotify.com/my-applications/#!/applications to generate your own'
```
Zarthus: Same as above.

Zarthus:
I think if you're going to be reformatting code it would be better to do it in separate commits. It's a pain to review otherwise.

butlerx:
sorry I didn't mean to actually reformat the code. my editor had a autopep8 function i didnt notice was run

flashcode:
Merged, thanks.

PR #244, opened 2017-11-10
autosort.py 3.1: Use colors to format the help text.

de-vri-es:
This update uses colors to improve readability of the help text. It also tweaks the text itself a bit.


flashcode:
Merged, and bumped version.

PR #246, opened 2017-11-12
autosort.py 3.3: Fix debug command for unicode buffer names.

de-vri-es:
Changelog:
  * Fix the /autosort debug command for unicode.
  * Update the default rules to work better with Slack.

Sorry for the large number of PR's. 3.0 was a large change, so some problems were to be expected I suppose. I did test with both python 2 and 3 now.

### DIFF
--- a/python/autosort.py
+++ b/python/autosort.py
@@ -25,6 +25,9 @@
 
 #
 # Changelog:
+# 3.3:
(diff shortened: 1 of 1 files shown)


de-vri-es:
Actually, the slack default may not be the best idea. Closing this to think a bit more first.

PR #247, opened 2017-11-12
autosort.py 3.3: Fix debug command for unicode buffer names.

de-vri-es:
Changelog:
  * Fix the /autosort debug command for unicode.
  * Update the default rules to work better with Slack.

I updated the defaults since #246 after some more discussion on IRC (couldn't re-open the other PR because I force-pushed to the branch).

### DIFF
--- a/python/autosort.py
+++ b/python/autosort.py
@@ -25,6 +25,9 @@
 
 #
 # Changelog:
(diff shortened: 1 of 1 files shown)


de-vri-es:
thanks!

PR #252, opened 2018-02-15
triggerreply.py 0.3: added regex support in triggers

vlad-s:
triggerreply.py v0.3 - now with regex support in triggers.

15-02-2018 - Vlad Stoica
Added regex support in triggers, and edited syntax of adding triggers.
The command is now 'add "trigger" "reply"'. Quote marks can be escaped
in triggers or replies by prefixing them with a backslash ('\\'). For
example, 'add "\\"picture\\"" "say \\"cheese\\"!"' is a valid command and
will reply with 'say "cheese"!' whenever it finds '"picture"' sent.

### DIFF
--- a/python/triggerreply.py
+++ b/python/triggerreply.py
@@ -102,10 +102,13 @@ def search_trig_cb(data, buf, date, tags, displayed, highlight, prefix, message)
             return weechat.WEECHAT_RC_OK
 
     for row in cursor.execute("SELECT * FROM triggers"):
-        r = re.compile(row[1])
-        if r.search(message) is not None:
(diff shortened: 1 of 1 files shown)


weechatter:
Why not using the trigger plugin? in my opinion this script is outdated

vlad-s:
After more than two years of last updating the plugin, I received an email with a question about it, regarding regex support. I could help, so I did. If the plugin is outdated, I have no problem with it being removed from the repo. It's just a simple way to add an automated reply to a trigger.

flashcode:
Merged, thanks.

Issue #197, opened 2017-03-12
colorize_nicks.py: Freezes weechat with too many nicks on one line


xnrand:
When `colorize_nicks.look.greedy_matching` is enabled, sending a line with the same nickname repeated many times freezes weechat.

**Workaround:** `/set colorize_nicks.look.greedy_matching off`

Channel log:
```
Mar12[19:19:18] <+Zzzz> Zzzz Zzzz Zzzz Zzzz Zzzz Zzzz Zzzz Zzzz Zzzz Zzzz Zzzz Zzzz Zzzz Zzzz Zzzz Zzzz Zzzz Zzzz Zzzz Zzzz Zzzz Zzzz Zzzz Zzzz Zzzz Zzzz Zzzz Zzzz Zzzz Zzzz Zzzz Zzzz Zzzz Zzzz Zzzz Zzzz Zzzz Zzzz Zzzz Zzzz Zzzz Zzzz Zzzz Zzzz Zzzz Zzzz Zzzz Zzzz Zzzz Zzzz Zzzz Zzzz Zzzz Zzzz Zzzz Zzzz Zzzz Zzzz Zzzz Zzzz Zzzz Zzzz Zzzz Zzzz Zzzz Zzzz Zzzz Zzzz Zzzz Zzzz Zzzz Zzzz Zzzz Zzzz Zzzz Zzzz Zzzz Zzzz Zzzz Zzzz Zzzz
Mar12[19:19:18] <+Zzzz> Zzzz Zzzz Zzzz Zzzz Zzzz Zzzz Zzzz Zzzz Zzzz Zzzz Zzzz Zzzz Zzzz Zzzz Zzzz Zzzz Zzzz Zzzz Zzzz Zzzz Zzzz Zzzz Zzzz Zzzz Zzzz Zzzz Zzzz Zzzz Zzzz Zzzz Zzzz Zzzz Zzzz Zzzz Zzzz Zzzz Zzzz Zzzz Zzzz Zzzz Zzzz Zzzz Zzzz Zzzz Zzzz Zzzz Zzzz Zzzz Zzzz Zzzz Zzzz Zzzz Zzzz Zzzz Zzzz Zzzz Zzzz Zzzz Zzzz Zzzz Zzzz Zzzz Zzzz Zzzz Zzzz Zzzz Zzzz Zzzz Zzzz 
Mar12[19:23:09] <-- testclient (~grumble@[host omitted for this github issue]) has quit (Ping timeout: 240 seconds)
```
[...]

PR #259, opened 2018-03-11
colorize_nicks.py: fix freezes from too many nicks in one line

alyptik:
With `greedy_matching` enabled, too many nicks on one
line completely locks up weechat.

Add a match_limit` configuration variable (defaults to 20)
which is used as a hard limit for number of greedy matches
allowed, at which point the plugin will fall back to the lazy
matching algorithm.

Fixes issue #197.


flashcode:
Merged, thanks.

PR #261, opened 2018-03-26
Fix/add feautures for mnotify.py

Tony763:
Hello,
I fixed some isues with dcc not working properly and added som features. I sended it to an author, but he is not responding.

Fix: Changed dcc redex to match new notify appears (weechat notify now contain IP) - only ddc get request worked
Fix: Dcc send info about file even with spaces in filename.
Feature: DCC get request show name with ip, network and size of file.
Feature: Added dcc send offer and dcc send start notify
Feature: Setting for notify is divided to off(don't send), on(always send), away(send only when away).
Changed default settings to match new scheme
Feature: Added small help - I get hard time to find out why no message is send
All redex checked on https://pythex.org/

### DIFF
--- a/python/mnotify.py
+++ b/python/mnotify.py
@@ -10,7 +10,24 @@
 # based on:
 # growl.py
 # Copyright (c) 2011 Sorin Ionescu <sorin.ionescu@gmail.com>
-
+#
+#Changelog:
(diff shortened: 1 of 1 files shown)


flashcode:
Thanks, merged (I fixed some style issues).

Issue #163, opened 2016-07-07
buffer_autoset.py 0.9: apply properties to already opened buffers


mumixam:
A user was trying to use buffer_autoset to hide the buffer highmon from highmon.pl, but this doesnt work because the highmon buffer is opened before buffer_autoset.py is loaded. I found this as a workaround `/set weechat.startup.command_after_plugins "/perl reload highmon"`

sim642:
As of version buffer_autoset.py 0.9 it does instantly apply options to buffers which are already open and match. I think the issue, however, is that it doesn't do the instant application procedure when the script and its configuration are loaded.

mumixam:
Yea correct, if you reload highmon, or if you set the buffer property to hide 1 again it becomes hidden, my thoughts are the perl plugin gets loaded before the python one so highmon's buffer is opened before buffer_autoset is loaded. I tested this with latest version of buffer_autoset and 1.5 stable weechat

flashcode:
That would be a reason to move this buffer_autoset script into core or as a C plugin. This way it can be loaded before anything else.

PR #264, opened 2018-04-14
buffer_autoset.py: on startup apply settings to already opened buffers

kimheino:
For example jabber buffers are already opened when buffer_autoset.py
runs. This should fix issue #163.


flashcode:
Merged and bumped the version, thanks!

PR #267, opened 2018-06-02
Update URLs from "http://www.weechat.org" to "https://weechat.org"

Zarthus:
Also bump all versions by its lowest value + 1

Possibly affects with a breaking change (doubt it): WeeGet

All of it was automated, with some manual corrections.

Conflicts with any open PR that updates one of these files, but the merge conflicts should be abysmal to fix.

Also replaces http://weechat.org to https://weechat.org

### DIFF
--- a/lua/http_item.lua
+++ b/lua/http_item.lua
@@ -26,14 +26,14 @@ local w = weechat
 
 SCRIPT_NAME    = "http_item"
 SCRIPT_AUTHOR  = "xt <xt@bash.no>"
-SCRIPT_VERSION = "1"
+SCRIPT_VERSION = "2"
 SCRIPT_LICENSE = "GPL3"
 SCRIPT_DESC    = "Bar item with HTTP source, using lua patterns to match content"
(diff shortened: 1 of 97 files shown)


flashcode:
Thanks for the PR.
For now, I consider it's not a major problem to have an old URL in scripts: it's always in comments or messages displayed, and there's an automatic redirection to the new URL.
Bumping all scripts versions at once for just an URL changed in a message is not really needed now.
I'll take care about scripts updates and change the URL each time a script is updated, if it points to an old URL.

de-vri-es:
For a change this minor, it might even be an option to just change the URL and not bump the version? I presume nothing would break horribly.

Zarthus:
I proposed the same, but FlashCode didn't want to have a checksum change without version bump.

flashcode:
Yes, the script plugin checks new version with the checksum of the file: so if only one byte in the file has changed, the script is marked as obsolete in WeeChat (which means a new version can be instaled). The comparison is not made on the script version.

de-vri-es:
Fair enough. Good day to you both :)

PR #273, opened 2018-06-19
aformat.py 0.2: make it py3k-ok

Hairo:
Tested with [weechat-python3-git](https://aur.archlinux.org/packages/weechat-python3-git/) from AUR.

### DIFF
--- a/python/aformat.py
+++ b/python/aformat.py
@@ -23,6 +23,8 @@
 # History:
 #   2016-09-24:
 #       v0.1: Initial release
+#   2018-06-19:
+#       v0.2: py3k-ok
 #
 # TODO:
 # - Colors support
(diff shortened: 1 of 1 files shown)


python/aformat.py
```diff
@@ -34,13 +36,13 @@
     from weechat import WEECHAT_RC_OK
     import_ok = True
 except ImportError:
-    print "This script must be run under WeeChat."
-    print "Get WeeChat now at: http://www.weechat.org/"
+    print("This script must be run under WeeChat.")
```
sim642: I think this requires `from __future__ import print_function` on top of the file to make the script still work under Python 2 as well, since it doesn't have the function by default.

Hairo: As long as you're not using `print("str1", "str2")` it will still print as intended, because it's the same as `print ("str")` (still a valid statement and not a funcion in py2.x). 

See: https://stackoverflow.com/questions/12162629/
and an example: https://ideone.com/KWqSle

PR #275, opened 2018-06-21
apply_corrections 1.3: make py3-ok, readd bugfix

raugturi:
- Update to support Python 3
- It appears the fix for one of the variables being used before it was declared got removed by mistake in the last release. This adds that change back.

### DIFF
--- a/python/apply_corrections.py
+++ b/python/apply_corrections.py
@@ -58,9 +58,14 @@
 
 # History:
 #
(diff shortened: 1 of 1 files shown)


flashcode:
Merged, thanks.
I think nothing was removed by mistake, but I made changes in the script on hook_print, and I think you didn't submitted your version 1.2.

raugturi:
@flashcode Oops. Sorry about that. Either way it should all be in there now.

PR #277, opened 2018-07-03
[WIP] notify.py 0.0.9 Works with weechat compiled with python3 now

ktsamis:
Modified the notify.py script to work with python3 compiled weechat. The script still needs the python2-notify2 package but it works with weechat 2.1 compiled with python3 now. I don't know if it should be tagged as py3k-ok yet since it depends on python2 package.

### DIFF
--- a/python/notify.py
+++ b/python/notify.py
@@ -38,6 +38,7 @@
 #     version 0.0.3: added config options for icon and urgency
 # 2009-05-02, FlashCode <flashcode@flashtux.org>:
 #     version 0.0.2.1: sync with last API changes
+from __future__ import print_function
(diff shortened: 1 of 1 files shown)


ktsamis:
Nope, this version of the script causes weechat to crash. Don't merge yet.

flashcode:
Hi,
Is this still in progress?
Once done, please rebase all your commits in a single one and use the commit message described in the Contributing file.
Thanks.

ktsamis:
I haven't actually made this work without causing the SIGSEV, I will close this, I am using another notification script and will not do any more work on this.

PR #278, opened 2018-07-07
jump_smart_higher.pl 0.3: reverse direction jumping, new commands /jump_smart_previous and /jump_smart_next, renamed from jump_smart_higher.pl to jump_smart_closest.pl

arza-zara:


flashcode:
@arza-zara: I think you forgot to rename the script in the call to `weechat::register` function.

arza-zara:
Fixed.

PR #284, opened 2018-08-20
weetweet.py 1.2.8: Do not use the deprecated user streams anymore

DarkDefender:
The diff is so huge because of some general code style cleanups.

However, the only real change should be that weetweet now doesn't use the depreceated twitter API anymore.


DarkDefender:
I've now updated to version 1.2.9. It would be nice if these changes could be merged in the script repo... :P

PR #285, opened 2018-08-20
pyrnotify.py: make it work with python3

jpmenil:
Still need to be sure that it is python2 compatible too.


jpmenil:
Work also in python2

flashcode:
Merged, thanks.

I added the missing import of `print_function` from `__future__`.

PR #292, opened 2018-10-02
weemustfeed.py: Py3ok and various fixes.

MacGyverNL:
Make Python3 compatible.
Feed fetching should now timeout after 15 seconds.
Handle negative return values correctly.
Fix deleting from active iteration object in unload.

Tested on 2.2 with python 2.7.15 and on 2.3-dev with python 3.7.0


flashcode:
Merged, thanks.

PR #293, opened 2018-10-03
pybuffer.py: Python3 compatibility

MacGyverNL:
Tested on 2.2 with Python 2.7.15, and on 2.3-dev with Python 3.7.0

Tests consisted of playing around with the python interpreter provided
in /pybuffer, as well as adding as a module to colorize_nicks.py and
playing around in that debug-buffer.

I *think* this is sufficient, although futurize suggests importing their "from builtins import object". I don't think this is required because the objects of this library are not iterated over, nor converted to strings, etc.
I don't know what happens if one would try to use the python2.7 version to iterate over an object that does not have e.g. a next() method -- but such an object is itself not python2.7-compatible.

### DIFF
--- a/python/pybuffer.py
+++ b/python/pybuffer.py
@@ -40,7 +40,7 @@
 #   '0x9ca4ce0'
 #   >>> buffer_get(b, 'input')
 #   Traceback (most recent call last):
-#     File "<console>", line 1, in <module>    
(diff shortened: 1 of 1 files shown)


flashcode:
Merged, thanks (with update of WeeChat website URL).

PR #302, opened 2018-11-02
shortenurl.py 0.6.3: is.gd api call changed

jsaalfeld:
The api.php is depricated. So i changed it according to the
reference: https://is.gd/apishorteningreference.php

This is the current situation:

The change will result in the following situation:


weechatter:
is script python3 compatible?

jsaalfeld:
Since I just changed the call and tested it in my running `weechat` instance, I didn't make any tests on compatibility of python versions (additionally I don't know how to change the python version that `weechat` should use)

PR #307, opened 2018-12-08
buffer_autohide.py 0.3: Add new configuration option

notmatti:
This introduces a functionality to define excemptions for certain
buffers.


flashcode:
Thanks, merged.
I fixed a typo: "excemption" -> "exemption", and added Python 3 compatibility (just added import of the print function).

PR #309, opened 2018-12-21
lnotify.py 0.3.4: Redirect stderr to /dev/null, ignore urxvt

mkaito:
Sometimes `xdotool` has a hiccup. It may be unable to find the pid of a window, or maybe there's no window currently focused. When this happens, stderr is spit out into the terminal, garbling output and confusing ncurses permanently until both weechat and the terminal window are restarted. Given that these errors are completely harmless, ignore them by piping stderr into `/dev/null`.

Also, add `urxvt` to the list of ignored windows.

### DIFF
--- a/python/lnotify.py
+++ b/python/lnotify.py
@@ -42,6 +42,10 @@
 #
 # 0.3.3
 # Fix undefined ignore_windows_list.
+#
+# 0.3.4
+# - Redirect stderr when calling xdotool
+# - Add urxvt to the list of ignored windows
 
 import weechat as weechat
 import subprocess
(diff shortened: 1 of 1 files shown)


weechatter:
why not simply making "ignore_windows_list" as option?

mkaito:
Probably a good idea. My main focus here was to fix the `stderr` garbage, though.

mkaito:
So, should I make it an option for this PR before you merge?

weechatter:
yes. would be fine..

flashcode:
Still in progress, or should we close this MR?

PR #311, opened 2019-01-11
fullwidth.py 0.1.1: Correctly handle fullwidth spaces.

khyperia:
From the wikipedia article [Halfwidth and fullwidth forms](https://en.wikipedia.org/wiki/Halfwidth_and_fullwidth_forms):

> U+FF00 does not correspond to a fullwidth ASCII 20 (space character), since that role is already fulfilled by U+3000 "ideographic space".

(12288 is decimal U+3000)

---

Note: According to the [contributing guidelines](https://github.com/weechat/scripts/blob/master/Contributing.adoc), I have contacted `GermainZ <germanosz@gmail.com>` as of immediately after creating this PR. Hopefully they respond soon :)

Additionally, I checked their github profile page, and they don't seem to have a personal repository for fullwidth.py (it's... not a large script), so keeping code in sync shouldn't be a problem.

### DIFF
--- a/python/fullwidth.py
+++ b/python/fullwidth.py
@@ -27,7 +27,7 @@ import weechat
 
 SCRIPT_NAME = "fullwidth"
 SCRIPT_AUTHOR = "GermainZ <germanosz@gmail.com>"
-SCRIPT_VERSION = "0.1"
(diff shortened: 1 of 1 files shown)


GermainZ:
Looks good to me :)

PR #312, opened 2019-01-30
buffer_autohide 0.4: Optionally keep buffers active for a given time

notmatti:
This commit introduces the new configuration options `keep_open` and
`keep_open_timeout`. If `keep_open` is `off`, buffers that should get
hidden will be immediately hidden when switching buffers. Else, it will
be hidden after a certain timeout defined in `keep_open_timeout` has
been passed.

`keep_open_timeout` is given in milliseconds. It is possible to enter
the value as mathematical expression, i.e. `60 * 1000` for 60 milli-
seconds.

### DIFF
--- a/python/buffer_autohide.py
+++ b/python/buffer_autohide.py
@@ -1,7 +1,8 @@
 # -*- coding: utf-8 -*-
 # MIT License
 #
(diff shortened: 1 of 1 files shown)


notmatti:
Closing in favor of #314

PR #315, opened 2019-02-08
autopong: Only reply to public messages that only contain the word ping

evgeni:
not words ending in "ping" or sentences ending in "ping"


flashcode:
Thanks, merged.

PR #316, opened 2019-02-09
url_arza.pl 0.2: support multiple urls per line, fix default values, fix truncated prefix, increase timeout, simplify

arza-zara:


weechatter:
why not making the $timeout variable customizable arza? ;-)

arza-zara:
I don't think it's necessary.

PR #317, opened 2019-02-20
shortenurl.py 0.6.4: add user agent to fix 403 on is.gd

jsaalfeld:
It seems that is.gd updated their policy and a 'User-Agent' is needed on requests. I shortly checked that with the following script:

```
import urllib2

string = "http://is.gd/api.php?longurl=https%3A%2F%2Fmedia.ccc.de%2Fv%2Fbiometrie-s8-iris "

req = urllib2.Request(string)
try:
    print("without agent:")
    res = urllib2.urlopen(req)
    ans = res.read()
    print(ans)
except Exception, e:
    print(e)

opener = urllib2.build_opener()
opener.addheaders = [('User-Agent', 'weechat')]
try:
    print("with agent:")
    resp = opener.open(string).read()
    print(resp)
except Exception, e:
    print(e)
```

which results in
```
without agent:
HTTP Error 403: Forbidden
with agent:
https://is.gd/Y3BSrS
```

And updated the plugin accordingly.

### DIFF
--- a/python/shortenurl.py
+++ b/python/shortenurl.py
@@ -14,6 +14,9 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 # History
+# 2019-02-20, Jochen Saalfeld <privat@jochen-saalfeld.de>
+#  version 0.6.4: Fix is.gd URL pulling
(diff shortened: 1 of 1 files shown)


flashcode:
Thanks.

PR #327, opened 2019-03-24
Fix email that is no longer reachable

kevr:
… to one that is and shall be indefinitely (until google goes down)

Original author of script. Just in case anybody wanted to contact me in the future; `kevr@nixcode.us` has been down for quite a while now.

Signed-off-by: Kevin Morris <kevr.gtalk@gmail.com>


flashcode:
Thanks, merged.

PR #328, opened 2019-03-26
mnotify.py: Python 2 and 3 support

Tony763:


flashcode:
Thanks, merged without some changes not needed.

PR #331, opened 2019-04-05
autosort 3.4: Fix rate limiting and empty argument for info hook.

de-vri-es:
Changelog:
  * Fix rate-limit of sorting to prevent high CPU load and lock-ups.
  * Fix bug in parsing empty arguments for info hooks.
  * Add debug_log option to aid with debugging.
  * Correct a few typos.

### DIFF
--- a/python/autosort.py
+++ b/python/autosort.py
@@ -25,6 +25,11 @@
 
 #
 # Changelog:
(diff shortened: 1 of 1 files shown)


de-vri-es:
Thanks!

PR #335, opened 2019-05-03
lnotify.py 0.35: fix malformed `ps` call with py3

### LINKED ISSUE #334: lnotify 0.34 + Python 3.7: malformed ps call
`subprocess.check_output()` returns `bytes`.
The result of the call on line 103 (bytes) is then used to form the command passed to another call of `subprocess.check_output()` on the following line.

This fails on Python 3.

When lnotify is run on Python 2, the shell command on line 104 is obtained by using `%` string formatting (a Python 2 default), with a `bytes` object used as the value of the singular `%s`. This provides a clean output in Python 2 (say, since it's a PID, `1102`) and a mangled one in Python 3 : `b'1102'`, since there is no implicit bytes -> str conversion (apart from bytes' `__str__`).

`window_pid`, the `bytes` instance, must be converted to `str` (unicode string) in Python 3, before using it as an argument to `%`.

I'll submit a PR fixing this using `from __future__ import unicode_literals`.

randrej:
This fixes #334. Please, test it on Python 2, too.

### DIFF
--- a/python/lnotify.py
+++ b/python/lnotify.py
@@ -42,6 +42,11 @@
 #
 # 0.3.3
 # Fix undefined ignore_windows_list.
+#
+# 0.3.5
+# Fix ps call generation in Python 3
+
+from __future__ import unicode_literals
 
 import weechat as weechat
 import subprocess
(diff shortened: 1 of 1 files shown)


weechatter:
please bump version number:
lnotify_version = "0.3.4"

flashcode:
And please use the commit message mentioned in the file [Contributing.adoc](https://github.com/weechat/scripts/blob/master/Contributing.adoc).

randrej:
Changed title + version.

sadsfae:
This works for me, tested on Python 3.7.3

flashcode:
Thanks, merged.

PR #338, opened 2019-05-13
highmon.pl v2.6: improvement - send logger_backlog signal if logging enabled

HubbeKing:
Script author hasn't done anything on github since 2014, thought I'd fix this thing which'd been bugging me for months.

highmon buffer now sends a "logger_backlog" signal once the buffer has been opened, if logging is enabled and the logger plugin appears to be loaded. 
This makes highmon.pl more usable in situations where one needs to frequently restart weechat.

### DIFF
--- a/perl/highmon.pl
+++ b/perl/highmon.pl
@@ -322,11 +322,7 @@ sub highmon_buffer_open
 		# send "logger_backlog" signal if logging is enabled to display backlog
 		if (weechat::config_get_plugin("logging") eq "on")
 		{
-			# this should be true only if logger plugin is loaded and auto_log is enabled
-			if (weechat::config_get("logger.file.auto_log"))
-			{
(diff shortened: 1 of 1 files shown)


perl/highmon.pl
```diff
@@ -317,6 +319,15 @@ sub highmon_buffer_open
 		{
 			weechat::buffer_set($highmon_buffer, "localvar_set_no_log", "1");
 		}
+		# send "logger_backlog" signal if logging is enabled to display backlog
+		if (weechat::config_get_plugin("logging") eq "on")
```
flashcode: Hi,
You don't have to check if the log is enabled, the logger_backlog callback function already does that.
You can just send the signal in all cases, the logger plugin is OK with that.

flashcode:
Merged, thanks.

PR #342, opened 2019-06-07
grep_filter.py 0.10: Remove newlines from command completion

trygveaa:
The enable and disable commands included newlines in the completion
template. When multiline.pl is loaded, this caused newlines to be shown
in the input bar after completing those commands.

I didn't find any other repo for this script. Since the author hasn't updated the script in three years, and this is such a small change, I didn't contact the author about it.

### DIFF
--- a/python/grep_filter.py
+++ b/python/grep_filter.py
@@ -19,6 +19,8 @@
 #
 # History:
 #
+# 2019-06-07, Trygve Aaberge <trygveaa@gmail.com>
+#   version 0.10: remove newlines from command completion
 # 2015-10-04, Simmo Saan <simmo.saan@gmail.com>
 #   version 0.9: fix text search imitation in filter
 # 2015-08-27, Simmo Saan <simmo.saan@gmail.com>
@@ -47,7 +49,7 @@ from __future__ import print_function
 
 SCRIPT_NAME = "grep_filter"
 SCRIPT_AUTHOR = "Simmo Saan <simmo.saan@gmail.com>"
-SCRIPT_VERSION = "0.9"
+SCRIPT_VERSION = "0.10"
 SCRIPT_LICENSE = "GPL3"
 SCRIPT_DESC = "Filter buffers automatically while searching them"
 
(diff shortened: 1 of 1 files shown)


sim642:
> I didn't find any other repo for this script.

Ummm, there's a repository URL given in the script for this exact reason, four lines below the version number you changed...
https://github.com/weechat/scripts/blob/6308382a2e365f0d7440c4ab0e2c995d7578c8a8/python/grep_filter.py#L56

And the script is in no way dead: the just-released WeeChat 2.5 includes an optimization specifically made to improve the usability of this script (https://github.com/weechat/weechat/pull/1311).

Futhermore, it seems to me like this fix should be in multiline.pl instead if it's affected by the formatting of the completion string in the script source code (spreading the completion string over multiple lines for readability is in no way limited to just this script). The presence of a newline in the completion string is completely fine for all WeeChat use without multiline.pl (it does not cause the completion to add a newline, which in normal operation would send the message).

Or if completion strings shouldn't actually contain extra whitespace (including but not limited to newlines), then that requirement should be documented in WeeChat's `hook_command` or be stripped in WeeChat itself. If extra spacing in the completion strings has no effect (?), then why would newlines be any different?

trygveaa:
> Ummm, there's a repository URL given in the script for this exact reason, four lines below the version number you changed...

Oh, my bad. For some reason, I didn't think to look for the repo url in the script when I didn't find it after searching for it, and I didn't notice it when changing the version number. I'm sorry.

> Futhermore, it seems to me like this fix should be in multiline.pl instead if it's affected by the formatting of the completion string in the script source code (spreading the completion string over multiple lines for readability is in no way limited to just this script). The presence of a newline in the completion string is completely fine for all WeeChat use without multiline.pl (it does not cause the completion to add a newline, which in normal operation would send the message).

The issue is not with multiline.pl, it's just more visible by it because it causes newlines in the input bar to be rendered as newlines, while weechat renders newlines in the input bar as a space by default. You can see that there's two characters (rendered as spaces) after the completion if you don't have multiline.pl installed, and if you press enter after completing it, you will send an empty message in addition to running the command (this happens regardless of if multiline.pl is installed or not).

> Or if completion strings shouldn't actually contain extra whitespace (including but not limited to newlines), then that requirement should be documented in WeeChat's hook_command or be stripped in WeeChat itself. If extra spacing in the completion strings has no effect (?), then why would newlines be any different?

Yeah, I agree that if there is no use for newlines in commands, weechat should strip them or replace them with spaces in the completion template.

flashcode:
FYI, there was a WeeChat issue for the newline char in completion, which I just fixed: https://github.com/weechat/weechat/issues/538

PR #346, opened 2019-06-17
highmon.pl v2.6 - Actually register new version nr

HubbeKing:
Turns out I was a bit of an idiot and forgot to update the register() function call with the new version number. Whoops.


flashcode:
Thanks, merged.

PR #352, opened 2019-07-11
go.py 2.6 & histsearch.py 0.4: fix detection of "/input search_text_here"

sim642:


flashcode:
Merged, thanks.
And I made histsearch.py compatible with Python 3.

PR #357, opened 2019-08-22
Updated autopong to Python3.

Heisenburg01:
Contacted Will Clouser - author.  OK


weechatter:
script is not needed anymore, you can easily use a /trigger to create this function:
https://github.com/weechat/weechat/wiki/Triggers#auto-respond-with-pong-to-ping-queries

Heisenburg01:
Should this script be removed from the repository?

weechatter:
i would recommend, yes.

Heisenburg01:
I removed autopong.py from the more recent commit.  If this is OK with everyone, then merge and close.  Thanks.

PR #360, opened 2019-09-10
Remove auto_away.py.  The same can be done with a trigger cmd_custom_…

Heisenburg01:
Removed auto_away.py.  The same can be done with a trigger cmd_custom_away in trigger examples.  Tried to contact author, no response.…away in trigger examples.  Tried to contact author, no response.

This is what I was able to figure out for this one.  If you think otherwise, please advise.


Heisenburg01:
Sorry if no change is wanted.  For some reason, I had this on my list of not upgraded to py3.  I see now I was wrong.

Heisenburg01:
Correction.  After review, I never attempted to contact this author.  I confused this with another script.
Please close as you see fit.

PR #361, opened 2019-09-13
ws_replace.py 0.1.1: fix python 3 compatibility

Streetwalrus:


weechatter:
for me the script is outdated. better use an /trigger:
for example: https://github.com/weechat/weechat/wiki/Triggers#remove-spaces-at-the-end-of-messagecommand-sent

Streetwalrus:
Thanks for the tip! Here's a trigger that matches the behavior of this script more closely:
`/trigger add spaces_out modifier "input_text_for_buffer" "" "/^\s*(.*)\s*$/${re:1}/"`

PR #363, opened 2019-09-21
topicdiff.py 0.4: make script compatible with Python 3

jspricke:


flashcode:
Hi,

I think the script will not work with Python 2 any more because `zip_longest` is not available. For now all scripts must be compatible with Python 2 and 3, during the transition.

I suggest to import with a try/except, like that:

```python
    try:
        from itertools import zip_longest
    except ImportError:
        from itertools import izip_longest as zip_longest
```

jspricke:
updated, thanks for the proposal.

PR #367, opened 2019-10-01
autoconnect.py 0.3.3: add python 3 compatinility

butlerx:
ive emailed arno as the original author for a review


flashcode:
Thanks, merged.

PR #370, opened 2019-10-02
away_action.py 0.6: make script compatible with python 3

quite:


quite:
I submitted a fix on the website before realizing that pushing it to the repo was better :)

PR #372, opened 2019-10-11
grep.py: fix a few more python3 errors

mweinelt:
```py3tb
Traceback (most recent call last):
  File "/var/lib/weechat/python/autoload/grep.py", line 1552, in cmd_logs
    msg = 'Found %s logs.' %len(file_list)
TypeError: object of type 'zip' has no len()
```

While `zip()` has a length in python2 it is lazy in python3 and needs to be evaluated first. Fix by evaluating the iterator before calculating it's length.

Calling list on a list in python2 is idempotent, so this keeps compatibility intact.

### DIFF
--- a/python/grep.py
+++ b/python/grep.py
@@ -596,7 +596,10 @@ def get_file_by_name(buffer_name):
             if '$server' in mask:
                 mask = mask.replace('$server', server)
         # change the unreplaced vars by '*'
-        from string import letters
+        try:
+            from string import letters
+        except ImportError:
+            from string import ascii_letters as letters
         if '%' in mask:
             # vars for time formatting
             mask = mask.replace('%', '$')
--- a/python/grep.py
+++ b/python/grep.py
@@ -1741,8 +1741,11 @@ Examples:
             debug = pybuffer.debugBuffer(globals(), '%s_debug' % SCRIPT_NAME)
         except:
             def debug(s, *args):
-                if not isinstance(s, basestring):
-                    s = str(s)
+                try:
+                    if not isinstance(s, basestring):
+                        s = str(s)
+                except NameError:
+                    pass
                 if args:
                     s = s %args
                 prnt('', '%s\t%s' %(script_nick, s))
--- a/python/grep.py
+++ b/python/grep.py
@@ -69,6 +69,9 @@
 #
 #   History:
 #
+#   2019-10-11, hexa-
+#   version 0.8.3: correct a few more python3 errors
+#
 #   2019-06-30, dabbill <dabbill@gmail.com>
 #               and Sébastien Helleu <flashcode@flashtux.org>
 #   version 0.8.2: make script compatible with Python 3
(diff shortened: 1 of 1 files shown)


mweinelt:
I'm still trying to track down a UnicodeDecodeError, so marked the pr as WIP.

mweinelt:
The UnicodeDecodeError is apparently related to https://github.com/weechat/weechat/issues/1389, so not WIP anymore.

flashcode:
@mweinelt: I just fixed the Python 3 issue, but this should then need another fix in the script. Now WeeChat will automatically send "bytes" instead of "str" to the callback if the string is not UTF-8 valid. See more info in the scripting guide here: https://weechat.org/files/doc/devel/weechat_scripting.en.html#_strings_received_in_callbacks

mweinelt:
Okay, will look into it.

flashcode:
@mweinelt: any news about the fix?

mweinelt:
This bug is quite infuriating. What is pickled here

https://github.com/weechat/scripts/blob/a8a4f17511ee589093af1e448607eae918dac2e1/python/grep.py#L1018

does not come out the same here

https://github.com/weechat/scripts/blob/a8a4f17511ee589093af1e448607eae918dac2e1/python/grep.py#L1047

In fact what was pickled cannot be properly decoded anymore. And the amounts of custom typing makes it hard to untangle.

koobs:
Came here via #377 and https://github.com/weechat/weechat/issues/1389 and  https://github.com/weechat/weechat/commit/513f5a1ee7ed88ee43b059827dca434edcc51e13 which may (appears to be) implicated in the following error seen in the following environment:

`
TypeError: can only concatenate str (not "bytes") to str
`
 
* Weechat: 2.7 (via FreeBSD Ports)
 * Python Plugin: Built against Python 3.8
 * Grep Script: 0.8.2

Full traceback: 

```
python: stdout/stderr (grep): Traceback (most recent call last):
python: stdout/stderr (grep):   File "/usr/home/user/.weechat/python/autoload/grep.py", line 1019, in grep_process_cb
python: stdout/stderr (grep):     grep_stdout += out
python: stdout/stderr (grep): TypeError: can only concatenate str (not "bytes") to str
‼ python: error in function "grep_process_cb"
```

Perhaps relevant to resolution, Voldenet (via IRC) mentioned:

```
Voldenet | grep_stdout += out.decode()
```

Since this is an existing (and open) PR addressing the above issue as well, I reported it here. If you'd like it in a separate issue (or reopening https://github.com/weechat/scripts/issues/377) let me know

voldenet:
Actually, after reading some more docs, `.decode` is useful only if output is not proper utf8 and can be consumed as bytes. In this case, it mostly can't.

The problem stems from printing wrong utf8 to stdout (for some reason). I think it would be best if grep_file opened files as utf8, since irc logs can contain random junk (not just valid utf8).

I did not reproduce the problem, but what I guess might help would be replacing `open(file, 'r')` with
```python
open(file, 'r', encoding='utf-8', errors='ignore')
```

koobs:
This is probably the better/best approach. Many Python package setup.py's have the same issue reading non-ASCII content to populate setup.py:description [1]

[1] https://github.com/rbarrois/python-semanticversion/commit/79fb67f86c73092ac79dc0d2d371b9a6ab6adb79

flashcode:
Closed in favor of #408.

PR #373, opened 2019-10-13
twitch.py 0.8: updated script

mumixam:
changed input modifier hooks to use irc_in2_* instead to avoid py3 unicode decoding errors
added setting 'plugins.var.python.twitch.notice_notify_block'
added setting 'plugins.var.python.twitch.client_id'

@flashcode you can bump min version again for this script up to 2.3 as i added 'weechat.hook_line' to it


flashcode:
Thanks, merged, and min WeeChat version has been set to 2.3.

PR #381, opened 2019-11-05
ripgrep.py: init at 0.1

mweinelt:
Adds ripgrep.py, a no frills grep module based on ripgreps --json output.

It still has some quirks here and there and is not very feature rich, but that is enough for me personally. It currently looks like this:

2019-11-05-204343_1505x533_scrot

It's not yet finished though, but I'd like to ask for feedback at this point.

TODO:
- [ ] License header
- [ ] Changelog
- [ ] Settings
  - Path to ripgrep binary
- [ ] ripgrep version check (>=0.10.0)
- [ ] Hook command description
… and other stuff I keep forgetting.

At one point I've seen it leak memory, but I currently don't seem to be able to reproduce that.

It is currently python 3.6+ only due to the use of f-strings. The ripgrep version required is packaged starting Debian 10 and Ubuntu 19.04, which also feature a recent enough python version.

Feedback very welcome.

### DIFF
--- a/python/ripgrep.py
+++ b/python/ripgrep.py
@@ -1,0 +1,280 @@
+import json
+import time
+
+try:
+    import weechat
+except ImportError:
+    weechat = None
+    print("This script requires WeeChat to run.")
+
+SCRIPT_NAME = "ripgrep"
+SCRIPT_AUTHOR = "Martin Weinelt <martin+weechat@linuxlounge.net>"
+SCRIPT_VERSION = "0.1"
+SCRIPT_LICENSE = "GPL3"
+SCRIPT_DESCRIPTION = "Use /rg to search through buffer logfiles"
+SCRIPT_SHUTDOWN_FUNC = ""
+SCRIPT_CHARSET = ""
+
+STATE = {"BUSY": False, "BUFFER": list()}
+CONFIG = {}
+SETTINGS = {"ripgrep_path": "/usr/bin/env rg"}
+
+
+def load_config_values() -> None:
+    CONFIG["WEECHAT_LOOK_PREFIX_JOIN"] = weechat.config_string(
+        weechat.config_get("weechat.look.prefix_join")
+    )
+    CONFIG["WEECHAT_LOOK_PREFIX_PART"] = weechat.config_string(
+        weechat.config_get("weechat.look.prefix_part")
+    )
+    CONFIG["WEECHAT_LOOK_PREFIX_NETWORK"] = weechat.config_string(
+        weechat.config_get("weechat.look.prefix_network")
+    )
+    CONFIG["WEECHAT_HISTORY_MAX_BUFFER_LINES_NUMBER"] = weechat.config_integer(
+        weechat.config_get("weechat.history.max_buffer_lines_number")
+    )
+    CONFIG["LOGGER_FILE_TIME_FORMAT"] = weechat.config_string(
+        weechat.config_get("logger.file.time_format")
+    )
+    CONFIG["IRC_COLOR_NICK_PREFIXES"] = weechat.config_string(
+        weechat.config_get("irc.color.nick_prefixes")
+    )
+
+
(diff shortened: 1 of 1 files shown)


python/ripgrep.py
```diff
@@ -0,0 +1,280 @@
...
+    # separate channel mode from nickname
+    try:
+        if nick[0] in "~@!%+":
+            nick = nick.lstrip("@!%+")
+    except IndexError:
+        pass
+
+    color_highlight = weechat.color("red")
+    color_default = weechat.color("chat")
+    color_nick = weechat.info_get("irc_nick_color", nick) or 0
```
nyuszika7h:
```suggestion
    color_nick = weechat.info_get("nick_color", nick) or 0
```

`irc_nick_color` is deprecated since WeeChat 1.5. (It still works, but it's probably preferable to change it.)

python/ripgrep.py
```diff
@@ -0,0 +1,280 @@
...
+    try:
+        color_nick_number = int(color_nick.replace("\x19F", "", 1)) or 0
```
nyuszika7h:
```suggestion
        color_nick_number = int(color_nick.replace("\x19F@", "", 1)) or 0
```

Without this change, I get an error on WeeChat 2.6:
```
13:25:30 python: stdout/stderr (ripgrep): Traceback (most recent call last):
13:25:30 python: stdout/stderr (ripgrep):   File "/home/nyuszika7h/.weechat/python/ripgrep.py", line 241, in rg_cb
13:25:30 python: stdout/stderr (ripgrep):     buffer_print_match(buffer, result)
13:25:30 python: stdout/stderr (ripgrep):   File "/home/nyuszika7h/.weechat/python/ripgrep.py", line 134, in buffer_print_match
13:25:30 python: stdout/stderr (ripgrep):     color_nick_number = int(color_nick.replace("\x19F", "", 1)) or 0
13:25:30 python: stdout/stderr (ripgrep): ValueError: invalid literal for int() with base 10: '@00130'
```

mweinelt: Try `/rg grep` in #weechat. Fails for me like this with your change:

```py3tb
13:57 python: stdout/stderr (ripgrep): Traceback (most recent call last):
13:57 python: stdout/stderr (ripgrep):   File "/var/lib/weechat/python/ripgrep.py", line 241, in rg_cb
13:57 python: stdout/stderr (ripgrep):     buffer_print_match(buffer, result)
13:57 python: stdout/stderr (ripgrep):   File "/var/lib/weechat/python/ripgrep.py", line 134, in buffer_print_match
13:57 python: stdout/stderr (ripgrep):     color_nick_number = int(color_nick.replace("\x19F@", "", 1)) or 0
13:57 python: stdout/stderr (ripgrep): ValueError: invalid literal for int() with base 10: '\x19F11'
13:57 ⚠  python: error in function "rg_cb"
```

I guess I have to read up on the documentation of color codes in weechat. The currently implementation is naïve in that regard.

nyuszika7h:
```suggestion
        color_nick_number = int(color_nick.replace("\x19F", "", 1).lstrip("@")) or 0
```

Seems like some nick colors are like `@00062` while others are just like `10`. This should do the trick.

nyuszika7h:
The highlighting seems a bit buggy, I did `/rg nils_2.*weechat` in `#weechat` and in some places part of the prefix (nick) was highlighted but none of the actual message was.

image

python/ripgrep.py
mweinelt: Applied, thanks!

python/ripgrep.py
mweinelt: Applied, thank you!

flashcode:
@mweinelt: please use a separate repository for development of scripts, this repository is just the image of scripts hosted on weechat.org, new scripts can not be submitted via pull requests (only via https://weechat.org/scripts/add).

PR #385, opened 2019-12-07
Update listbuffer.py

ElonSatoshi:
Fixed version and link to latest development version

### DIFF
--- a/python/listbuffer.py
+++ b/python/listbuffer.py
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 #
-# ListBuffer, version 0.8.1 for WeeChat version 0.3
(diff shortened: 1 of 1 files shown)


flashcode:
Hi,

Thanks for the PR, but it will not be merged, for two reasons:

1. The typo on script version in comment will lead to a file checksum change, and WeeChat computes the checksum to know if a new script version is available. So changing a single char in a comment should lead to a new script version and will be shown for users as "new version". So I think it's better to make that change on the next real update on the script.

2. The URL to the development repository is already OK, and should be updated by the author only. This repository (weechat/scripts) is just a snapshot of user's scripts, it is NOT used for developments.

PR #386, opened 2019-12-19
adapt fileaway.py for Python 3

aperum:
This will make this script work with Python 3. I'm no python programmer but i think the iter change makes this incompatible with Python 2.

### DIFF
--- a/python/fileaway.py
+++ b/python/fileaway.py
@@ -180,7 +180,7 @@ def check_timer():
 
 if __name__ == "__main__":
   if w.register(SCRIPT_NAME, SCRIPT_AUTHOR, SCRIPT_VERSION, SCRIPT_LICENSE, SCRIPT_DESC, "", ""):
-    for option, default_value in iter(settings.items()):
+    for option, default_value in settings.items():
       if not w.config_is_set_plugin(option):
         w.config_set_plugin(option, default_value)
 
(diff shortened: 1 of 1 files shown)


python/fileaway.py
```diff
@@ -178,11 +180,11 @@ def check_timer():
 
 if __name__ == "__main__":
   if w.register(SCRIPT_NAME, SCRIPT_AUTHOR, SCRIPT_VERSION, SCRIPT_LICENSE, SCRIPT_DESC, "", ""):
-    for option, default_value in settings.iteritems():
+    for option, default_value in iter(settings.items()):
```
flashcode:
```suggestion
    for option, default_value in settings.items():
```
Using `.items()` is enough here.

python/fileaway.py
```diff
@@ -72,19 +72,21 @@
 #  -Handles improper commands
 # Version 1.0.2 release - Jun 15, 2011
 #  -Added alternative for xset users (credit: sherpa9 at irc.freenode.net)
+# Version 1.0.3 release - Dec 19, 2019
+#  -Adapt for python3
```
flashcode: To use the print function in Python 2, you should add this import:

```python
from __future__ import print_function
```

aperum: added, tnx

flashcode:
Merged, thanks.

PR #392, opened 2020-01-23
Remove bugs with "whisper" commands

zxosazx:
Script v0.8 not work with incoming command "WHISPER" and not work if message no have tag "subscriber". I am not a programmer and speak English poorly, I want to apologize for the patch


weechatter:
What script you are talking about? Please contact script author.

PR #395, opened 2020-01-31
away-action.py 0.7: add include_text option

dargad:
I have added include_text option where forwarding of confidential messages is not allowed.
Also slightly re-factored command arguments formatting.


dargad:
Fixed according to your remarks.

PR #398, opened 2020-02-16
Fixed for Python 3+. Not tested if it still works for Python 2.

nikster2:
Since no one seemed to make an effort to make it work for newer Python I did my own try. Since I really don't understand the code the changes are based on 7 year old examples I found on Stackexchange and modified to work in fish.py :)

I've not tried if it still works in Python 2, so maybe someone could check.

I'm not really interested in doing more work on this. I just needed it for myself. So it works for me but won't give any support if it doesn't work for anyone else. I'm just sharing it if someone else has the same problems and would like to try my maybe working fix :)

### DIFF
--- a/python/fish.py
+++ b/python/fish.py
@@ -56,7 +56,7 @@ from __future__ import print_function
 
 SCRIPT_NAME = "fish"
 SCRIPT_AUTHOR = "David Flatz <david@upcs.at>"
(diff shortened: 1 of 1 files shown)


flashcode:
Closing in favor of #406, please comment on this PR if needed.
Thanks.

PR #399, opened 2020-02-25
buffer_dmenu.py 0.1.2: Add Python 3 support

Seirdy:
Previously, buffer_dmenu used `str.decode("utf-8")` in several places;
however, this is incompatible with Python 3 and unnecessary since Python
3 uses UTF-8 strings by default. Removing all uses of `.decode("utf-8")`
makes the script Python 3 compatible.

### DIFF
--- a/python/buffer_dmenu.py
+++ b/python/buffer_dmenu.py
@@ -24,6 +24,8 @@
 # Optionally requires i3-py [py2] (or i3ipc [py3]) to focus weechat in i3
 #
 # History:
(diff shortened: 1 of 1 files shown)


Seirdy:
Attempted to contact script author @Ferus as per Contributing.adoc; received email bounceback.

flashcode:
Hi,
Are you sure your changes are compatible with Python 2?
(compatibility with Python 2 is mandatory until May 1st, see: https://weechat.org/scripts/python3/).

PR #401, opened 2020-03-11
emoji_aliases.py 1.0.4: support last args with no " :" (PRIVMSG #c test:a:)

jesopo:
apologies that I didn't catch this on the 1.0.3 fix but it occurred to me that not only does the RFC permit this, Oragono can actually do this.


flashcode:
Thanks, merged.

PR #402, opened 2020-03-12
make lastfm2.py compatible with python 3.x

repomaa:


flashcode:
Merged, with minor changes and edit of commit message.
Thanks.

PR #408, opened 2020-05-06
grep.py 0.8.3: more python3 compatibility fixes

### LINKED ISSUE #404: grep.py: tries to concatenate bytes in strings in python3
## Script / WeeChat

- Name of script: grep.py
- Script version: 0.8.2
- WeeChat version: 2.7.1

This bug has been reported to the script author? Yes/No Yes
cc @m4v 
Response or reason: 

Not yet responded just mentioned in this issue

## Bug summary

07:10 python: registered script "grep", version 0.8.2 (Search in buffers and logs)
07:10 python: stdout/stderr (grep): Traceback (most recent call last):
07:10 python: stdout/stderr (grep):   File "/home/joerg/.weechat/python/grep.py", line 1019, in grep_process_cb
07:10 python: stdout/stderr (grep):     grep_stdout += out
07:10 python: stdout/stderr (grep): TypeError: can only concatenate str (not "bytes") to str

## Steps to reproduce

1. /script install grep.py
2. /grep foo

## Current behavior

raises an exception

## Expected behavior

returns a result

## Suggested solutions

It either uses byte strings or strings consistently

martinetd:
- use pickle.dumps(s, 0) to force older ascii-compatible pickling protocol
This prevents encoding/decoding to/from utf8 to mess the string up
 - cast zip() to a list, as zip does not have length
 - cast range() to list, as range cannot reverse()
 - open file in utf8 and ignore error just in case

Fixes #404

---------

Replace #372 that has been inactive for ages.

This appears to work for me, but might be missing more - I tried normal grep, count, context, /logs... at least what I'm using is now functional.
Honestly the whole pickling doesn't make sense to me, I'm not sure what the point was, but weechat apparently does unholy things which corrupt the string (doing the same without specifying version 0 of the protocol complains about truncated pickle?!)

I also didn't test in python2, but I believe I didn't do anything non-compatible... Did I mention I hate python? But this looked less work than finding an alternative before I started...

Anyway, thanks for maintaining!

### DIFF
--- a/python/grep.py
+++ b/python/grep.py
@@ -69,6 +69,9 @@
 #
 #   History:
 #
+#   2020-05-06, Dominique Martinet <asmadeus@codewreck.org> and hexa-
+#   version 0.8.3: more python3 compatibility fixes...
+#
 #   2019-06-30, dabbill <dabbill@gmail.com>
 #               and Sébastien Helleu <flashcode@flashtux.org>
 #   version 0.8.2: make script compatible with Python 3
@@ -230,7 +233,7 @@ except ImportError:
 
 SCRIPT_NAME    = "grep"
 SCRIPT_AUTHOR  = "Elián Hanisch <lambdae2@gmail.com>"
-SCRIPT_VERSION = "0.8.2"
+SCRIPT_VERSION = "0.8.3"
 SCRIPT_LICENSE = "GPL3"
 SCRIPT_DESC    = "Search in buffers and logs"
 SCRIPT_COMMAND = "grep"
@@ -596,7 +599,10 @@ def get_file_by_name(buffer_name):
             if '$server' in mask:
                 mask = mask.replace('$server', server)
         # change the unreplaced vars by '*'
-        from string import letters
+        try:
+            from string import letters
+        except ImportError:
+            from string import ascii_letters as letters
         if '%' in mask:
             # vars for time formatting
             mask = mask.replace('%', '$')
@@ -984,7 +990,7 @@ def show_matching_lines():
             buffer_update()
         else:
             global hook_file_grep, grep_stdout, grep_stderr, pattern_tmpl
-            grep_stdout = grep_stderr = ''
+            grep_stdout = grep_stderr = b''
             hook_file_grep = weechat.hook_process(
                 'func:grep_process',
                 get_config_int('timeout_secs') * 1000,
@@ -1009,14 +1015,17 @@ def grep_process(*args):
     except Exception as e:
         result = e
 
-    return pickle.dumps(result)
-
-grep_stdout = grep_stderr = ''
+    return pickle.dumps(result, 0)
 
 def grep_process_cb(data, command, return_code, out, err):
     global grep_stdout, grep_stderr, matched_lines, hook_file_grep
 
+    if isinstance(out, str):
+        out = out.encode()
     grep_stdout += out
+
+    if isinstance(err, str):
+        err = err.encode()
     grep_stderr += err
 
     def set_buffer_error(message):
@@ -1548,7 +1557,7 @@ def cmd_logs(data, buffer, args):
     buffer = buffer_create()
     if get_config_boolean('clear_buffer'):
         weechat.buffer_clear(buffer)
-    file_list = zip(file_list, file_sizes)
+    file_list = list(zip(file_list, file_sizes))
     msg = 'Found %s logs.' %len(file_list)
 
     print_line(msg, buffer, display=True)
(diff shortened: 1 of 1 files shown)


flashcode:
Hi @martinetd,

Thanks for the PR.

There are also changes in WeeChat since version 2.7, about the strings sent to callbacks in Python, and this script must take care of that, for more info see the release notes and the scripting guide:

* https://weechat.org/files/releasenotes/ReleaseNotes-devel.html#v2.7_python3_callbacks_strings
* https://weechat.org/files/doc/stable/weechat_scripting.en.html#_strings_received_in_callbacks

I think the only way to test is to have some invalid UTF-8 data in logs, and then the script should receives bytes instead of str in the hook_process callback (when using Python 3).

martinetd:
> I think the only way to test is to have some invalid UTF-8 data in logs, and then the script should receives bytes instead of str in the hook_process callback (when using Python 3).

Pretty sure I ran into that... Tested on WeeChat 2.7.1.
This is what the "if isinstance(out, str)" were about initially, but converting back and forth invalid UTF-8 broke the pickle output hence forcing ASCII-armored output with the "version 0" option.
As written in the comment that whole pickling doesn't make much sense to me I have no idea what the original author was trying to do and will be blunt but already have spent too long to get here... I don't hate python but python3 with strings and bytes really made for a horrible experience :/

I'm not going to say there aren't any problems left, but this works for me.

python/grep.py
```diff
@@ -1009,14 +1012,17 @@ def grep_process(*args):
...
-    return pickle.dumps(result)
-
-grep_stdout = grep_stderr = ''
+    return pickle.dumps(result, 0)
 
 def grep_process_cb(data, command, return_code, out, err):
     global grep_stdout, grep_stderr, matched_lines, hook_file_grep
 
+    if isinstance(out, str):
+        out = out.encode()
```
flashcode, line 1021: I think you should do that only in Python 3 (same comment for the second encode below).
With Python 2, the type is always str and if there are unicode chars, this will raise a UnicodeDecodeError.

flashcode: By the way, Python 2 support is now optional, so we might drop Python 2 support now.
But since a lof of people is using the script, we could keep the compatibility for some time.

martinetd: Nah, I did test on python2 (well just the function, not whole script) -- string and byte alike will both match the isinstance but the .encode() will be no-op here as pickle with version 0 guarantees pure ascii. It should work.

I don't have any recent weechat with python2 to test... I can try on a way older weechat probably let me see

martinetd: confirmed, it works fine on weechat 2.3 with python2.

flashcode: Did you try with accents?

For example:

    Python 2.7.18 (default, Apr 20 2020, 20:30:41) 
    [GCC 9.3.0] on linux2
    Type "help", "copyright", "credits" or "license" for more information.
    >>> 'sébastien'
    's\xc3\xa9bastien'
    >>> 'sébastien'.encode()
    Traceback (most recent call last):
      File "<stdin>", line 1, in <module>
    UnicodeDecodeError: 'ascii' codec can't decode byte 0xc3 in position 1: ordinal not in range(128)

martinetd: There are weird characters in the log file I tried, but that doesn't matter - the only lines going in through that callback come from `pickle.dumps(result, 0)` which is guaranteed to be ascii. If that ever returns non-ascii then the python3 version won't work; I couldn't get the default pickle which isn't ascii (outputs invalid utf8) to work.

If you really don't want to call it in python2 I don't mind but I have no idea how to check current python version so you'll need to spoonfeed me that, I have no interest in spending the time to figure more python2/3 compat things out..

martinetd:
> Anyway, found another issue with grepping from buffer on the python3 side, will submit an update for that part..

Fixed that and credited @mweinelt as I stole the fix from his PR. Thanks!

martinetd:
(ah, actually now `/grep tést` in a buffer with tést in it will not match on python3, probably due to ascii_letters from the last update? But in log file it does work, and printing itself works fine; and I've just tried on python2 as well with same result and no change to that part of the code so I'm pretty sure it never worked so will just mark that as out of scope for this fix... Anyway I'm done with this, I never planned on spending so much time on that script, I'll do specific stuff you ask but no more from me.)

martinetd:
bump. Do you still need anything changed? I'd rather not let this rot in oblivion like the previous PR.

I explained why I didn't want to do the "requested change" (always encoding should be safe here and I prefer the (python2-only) performance cost to trying to fight with python to disable the behaviour on python2), if you disagree you'll need to ask again and I'll do it, otherwise I'd rather close the PR than let it stay open for months. Thanks.

flashcode:
Please be patient, I need time to validate and test pull requests, and I'm busy on other things for now.
By the way feedback or tests of changes from other people is welcome.
This will be merged soon, no need to close the PR.

martinetd:
Sorry if that sounded rude, I wasn't sure if you were still expecting a change or not.

PR #410, opened 2020-05-10
triggerreply.py 0.4.2: various improvements

fisherhu:
Added some new functions I needed.
Hopefully all documented.
Some optimizations still required.


weechatter:
for me the script is obsolete and can be removed from script repo, because /trigger plugin does the same, but better..

flashcode:
Squashed and merged, thanks!
Sorry for the delay…

PR #411, opened 2020-06-05
text_replace.py: add option to replace prefix only

normen:
As the title says, this way you can for example use a different character for commands (e.g. `:names` -> `/names`)

### DIFF
--- a/python/text_replace.py
+++ b/python/text_replace.py
@@ -20,6 +20,8 @@
 # (this script requires WeeChat 0.3.0 or newer)
 #
 # History:
+# 2020-06-05, Normen Hansen <normen667@gmail.com>
+#   version 0.7: add option to change prefix of entered text
 # 2019-06-17, Brad Hubbard <bhubbard@redhat.com>
 #   version 0.6: replace iteritems with items for python3 compatability
 # 2011-07-17, Sébastien Helleu <flashcode@flashtux.org>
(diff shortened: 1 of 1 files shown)


python/text_replace.py
```diff
@@ -38,14 +40,15 @@
...
-SCRIPT_VERSION = "0.6"
+SCRIPT_VERSION = "0.7"
 SCRIPT_LICENSE = "GPL3"
 SCRIPT_DESC    = "Replaces text you write with replacement text"
 
 # script options
 settings = {
         'replacement_pairs': '(:=:),):=:(',   # pairs separated by , orig text and replacement separated by =
         'replacement_words': 'hhe=heh',       # words separated by , orig text and replacement separated by =
+        'replacement_prefixes': ':=/',       # strings separated by , orig prefix and replacement separated by =
```
flashcode, line 51: Shouldn't this value be empty by default, so that there are no changes for existing users?
Not sure if this default value is OK for most users.

normen:
I don't know, if you want I can change it. I found it already quite cheeky that it replaces stuff by default so I thought ppl would configure the plugin anyway :)

flashcode:
OK, I'll merge as-is then.

flashcode:
Merged.

normen:
Cool, thanks!

PR #417, opened 2020-07-09
autojoinem.py 0.7: fix help to take advantage of evaluation in allchan

pascalpoitras:
fix help


weechatter:
thanks...

flashcode:
@pascalpoitras: you should mention that this example works only with Weechat ≥ 3.0.

weechatter:
flashcode, the brain :-) but its correct, maybe two examples.

pascalpoitras:
@flashcode done

pascalpoitras:
@weechatter i keep the original example and added the one that works only with ≥ 3.0.

flashcode:
By the way, I wonder if it's not the script job to do that, what about adding a command like `/autojoinem addall` that auto adds all opened channels?
And if done like that, the script can check itself if WeeChat version is ≥ 3.0 and use the evaluated version if needed.

PR #418, opened 2020-07-16
Port country.py to python3

TauPan:
The changes are identical to my previous PR, but for country.py only.

It sorta works for some users and fails for others.


flashcode:
Merged, thanks.

mumixam:
@flashcode Can you tag this as py3 on the weechat.org repo?

flashcode:
@mumixam: done.

PR #423, opened 2020-08-18
terminal_title.py 2: port to python3

alyssais:


alyssais:
Actually, it looks like there’s an upstream repository for this script. I’ll try fixing this there.

PR #420, opened 2020-08-02
fish.py: fix key exchange

wikro:
Exchanging keys with current version of fish.py (0.9.4) results in errors because sometimes bytes objects are passed to functions, sometimes string objects.

PR #425, opened 2020-09-02
Fix bug in key exchange again

bberna:
Using s.encode() for hashing the DH key is not correct
as the resulting bytearray is utf-8 encoded and shouldn't.
The resulting hash is then different from hashs produced by the
previous versions of this plugin, and probably other fish plugins (tested
with the original irssi one, and the key exchange didn't work either).
While using s.encode("latin-1") works, it's better to remove the encode() call
and rewrite int2bytes to actually return bytes (and not str). I used
python 3 built-ins for that, and modified bytes2int accordingly.

### DIFF
--- a/python/fish.py
+++ b/python/fish.py
@@ -61,7 +61,7 @@ from os import urandom
 
 SCRIPT_NAME = "fish"
 SCRIPT_AUTHOR = "David Flatz <david@upcs.at>"
-SCRIPT_VERSION = "0.9.5"
+SCRIPT_VERSION = "0.9.6"
 SCRIPT_LICENSE = "GPL3"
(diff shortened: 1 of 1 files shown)


flashcode:
Hi,
Is it related to #420?

bberna:
Yes, the problem with #420 is that the key exchange doesn't work as intended (the generated key is wrong, because encode() without argument default to "utf-8", while the correct codec to use here is "latin-1"). As explained, instead of encoding sha256 parameter, I modified bytes2int and int2bytes to actually manage "bytes" type instead of "str", as it should be in python3.

bberna:
Hi, do you need more informations about the fix before merging?

flashcode:
Merged, thanks.

PR #428, opened 2020-09-06
add autoaway2.py

jesopo:
I made this auto-away variant because:

1. I didn't want to be auto-unawayed
2. I wanted human interaction on weechat-android to count towards me not being away

as well as using `inactivity` to know when the user was last typing on weechat-ncurses, this script uses...

- sending a `PRIVMSG`
- being told we're no longer away - `RPL_UNAWAY` (305)
- buffers being marked as read (triggered when swiping/opening buffers in weechat-android)

... to determine unaway-ness

### DIFF
--- a/python/autoaway2.py
+++ b/python/autoaway2.py
@@ -1,0 +1,86 @@
+try:
+    import weechat as w
+    import_ok = True
+except ImportError:
+    print("This script must be run under WeeChat")
+    print("Get WeeChat now at: https://weechat.org/")
+    import_ok = False
+import time
+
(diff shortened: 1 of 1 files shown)


flashcode:
Hi,

Pull requests in this repository are for existing scripts only (see [Contributing.adoc](https://github.com/weechat/scripts/blob/master/Contributing.adoc)).
For new scripts, please use this form: https://weechat.org/scripts/add/

By the way, maybe you could add options to the existing script, or choose another name (same name with suffix 2 is not very clear about the differences…).

PR #422, opened 2020-08-17
colorize_nicks.py: fix nick colorization in input for colon completer

foxxx0:
The global `VALID_NICK` regex allows the nick to contain a colon, however
the nick matching logic with stripping the first/last character is not
used within `#colorize_input_cb()`.
In combination with a colon `:` for `weechat.completion.nick_completer`,
this leads to nicks not being colorized in the input line.

To fix this, we can just always `rstrip()` the currently configured
`nick_completer` character.

PR #435, opened 2020-09-30
colorize_nicks.py 28: fix highlighting of forms of address (nick: or @nick) for `colorize_input`

noctux:
This pullrequest addresses the same issue as #422 but provides additional mitigations:

Just a small recap on the issue: Newer versions of `colorize_nicks.py` have broken highlighting of nicks in the input-area when they are no exact matches, i.e. leading or trailing characters were added, as it it often occurs when using forms of address (@nick, nick:, etc.pp.). #422 tries to address this by specifically looking for the pattern 'nick${nick_completer}`. However, this bugfix is only partial, as it will not detect nicks in the following contexts:

```
nick1, nick2: nick1 does not end with nick_completor
@nick (even though I personally dislike this form of address, it happens at times)
nick!
Appositions in braces (see message by nick)
Ich mag nicks Aussage (German genitive, but yeah, it gets a bit exotic and out of scope here) 
```

For the chat area, all those statements highlight just fine, because the logic there drops a single character at the beginning as well as at the end (and the combination of both) and probes for nicks. This patch extracts that logic into a function and reuses it in the input callback.

I don't know what would be the best way to proceed from there: I'm happy to rebase on #422 once it is merged, however in my opinion it is sufficiently covered by this, more generic approach. Another option would be to move them behind an option, but to me that seems like unnecessary complexity from the user's perspective.

Please ping me on any potential issues!

(BTW: thanks for weechat, it's become an indispensable communication tool to me!)

Best regards,
Simon

### DIFF
--- a/python/colorize_nicks.py
+++ b/python/colorize_nicks.py
@@ -21,6 +21,9 @@
 #
 #
 # History:
+# 2020-09-05: noctux
+#   version 28: fix highlighting of forms of address (nick: or @nick)
+#               for `colorize_input`
 # 2020-05-09: Sébastien Helleu <flashcode@flashtux.org>
 #   version 27: add compatibility with new weechat_print modifier data
 #               (WeeChat >= 2.9)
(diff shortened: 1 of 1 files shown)


flashcode:
Hi,

I merged #443, so script version is now 28.
Moreover, PR #422 is also in progress, please wait for a merge of this PR before rebasing your changes on top of the latest version, and then you can bump version accordingly.

Thank you.

flashcode:
Hi,
A new version (29) has been uploaded, I close this pull request.
If needed, please open another one, based on the latest version of the script.
Thanks!

PR #436, opened 2020-10-01
grep.py: Fix Python 3 issue reversing range

thomwiggers:
`range` returns a generator in Python 3, and you can't call `.reverse()` on those.

### DIFF
--- a/python/grep.py
+++ b/python/grep.py
@@ -68,6 +68,8 @@
 #
 #
 #   History:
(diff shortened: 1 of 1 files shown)


flashcode:
Hi,
Could you please bump version in the script, add the fix to the changelog and format your commit message as described in the Contributing doc?
Thanks.

thomwiggers:
Ah, I'd missed that doc as Github doesn't autolink adoc...

flashcode:
Thanks!

PR #440, opened 2020-10-30
tmux_env.py 3.1: Fix compatibility problem with python 3.6 and kill python 2 support

TauPan:
I got the following backtrace with python 3.6 (which is still
supported and default on some distribution).

And testing with python2 seemed to complicated, so I simply removed
the text kwarg and called .decode() on the result.

See
https://docs.python.org/3.6/library/subprocess.html#subprocess.check_output

as opposed to:

https://docs.python.org/3.7/library/subprocess.html#subprocess.check_output

I don't know what would be necessary to still support python 2, also
that would entail support for 3 different python versions.

~~~python
08:29 python: stdout/stderr: (tmux_env): Traceback (most recent call last):
08:29 python: stdout/stderr: (tmux_env):   File "/home/delgado/.weechat/python/autoload/tmux_env.py", line 101, in timer_cb
08:29 python: stdout/stderr: (tmux_env):     update_environment()
08:29 python: stdout/stderr: (tmux_env):   File "/home/delgado/.weechat/python/autoload/tmux_env.py", line 119, in update_environment
08:29 python: stdout/stderr: (tmux_env):     env = subprocess.check_output(['tmux', 'showenv'], **check_output_kwargs)
08:29 python: stdout/stderr: (tmux_env):   File "/usr/lib64/python3.6/subprocess.py", line 356, in check_output
08:29 python: stdout/stderr: (tmux_env):     **kwargs).stdout
08:29 python: stdout/stderr: (tmux_env):   File "/usr/lib64/python3.6/subprocess.py", line 423, in run
08:29 python: stdout/stderr: (tmux_env):     with Popen(*popenargs, **kwargs) as process:
08:29 python: stdout/stderr: (tmux_env): TypeError: __init__() got an unexpected keyword argument 'text'
~~~

### DIFF
--- a/python/tmux_env.py
+++ b/python/tmux_env.py
@@ -24,6 +24,9 @@ History:
 
     2020-01-03 dobbymoodge <john.w.lamb [at] gmail . com> ( https://github.com/dobbymoodge/ )
       version 3: python 3.x compatibility
+
+    2020-10-30 Friedrich Delgado <taupan [at] gmail . com>
+      version 3.1: fix one missing instance of **check_output_kwargs
 """
 
 from __future__ import absolute_import, unicode_literals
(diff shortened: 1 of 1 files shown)


flashcode:
Thanks, merged.

Issue #72, opened 2015-05-28
urlgrab.py: Pressing enter / home / end does not work


matthijskooijman:
I just installed urlgrab and in the urlgrab buffer, pressing enter to open up an url didn't work. Running `/url **enter` works, so this is a problem with the keybinding. Pressing the up/down arrows worked normally, but home and end didn't work either.

It seems that there is an extra "meta" in the keybinding that my setup doesn't generate. Also, the home binding had an extra ./ that didn't happen here.

To fix this, I added these lines to the `init` function, and now all keys work:

        weechat.buffer_set(urlgrab_buffer, "key_bind_ctrl-J",        "/url **enter")
        weechat.buffer_set(urlgrab_buffer, "key_bind_ctrl-M",        "/url **enter")
        weechat.buffer_set(urlgrab_buffer, "key_bind_meta2-1~",    "/url **scroll_top")
        weechat.buffer_set(urlgrab_buffer, "key_bind_meta2-4~",      "/url **scroll_bottom")

Not sure if the existing bindings were wrong, or if some setups generate different sequences, but perhaps the above bindings should just be added? Or is something broken on my system?
I'm running weechat inside tmux inside uxterm, if that helps.

PR #441, opened 2020-11-08
urlgrab.py: change urlgrab buffer to open the URL with <Return>

Futurile:
Urlgrab.py adds some keyboard shortcuts, to open a URL when the user is in the buffer. The current behaviour is that you have to use <alt><return>, changed to be just plain <return>. The shortcut is limited to this buffer AFAICT. Also added it to the documentation as it doesn't tell you there are keyboard shortcuts.

There's an old issue #72 that mentions this

### DIFF
--- a/python/urlgrab.py
+++ b/python/urlgrab.py
@@ -652,7 +652,7 @@ def init():
         weechat.buffer_set(urlgrab_buffer, "key_bind_meta2-A",       "/url **up")
         weechat.buffer_set(urlgrab_buffer, "key_bind_meta2-B",       "/url **down")
         weechat.buffer_set(urlgrab_buffer, "key_bind_meta-ctrl-J",   "/url **enter")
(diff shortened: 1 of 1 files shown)


flashcode:
Hi,
I'm not sure if using Enter key is a good idea in any buffer, that means you disable the way for the user to enter a command in this buffer, is that really wanted?
All default plugins with a buffer (like fset, scripts, etc.) don't use Enter key to let the user use the command line line on other buffers.

flashcode:
If needed, maybe the script could provide a way to customize keys, with alt+Enter as default and the user could use Enter if needed.

Futurile:
Oh "gah!" on my part - I wondered when I saw the old request and was a
bit reticent as I thought there had to be some reason - been changing
buffer by mouse so didn't notice it.

flashcode:
Hi,
Are you working on updates?
If yes, please don't forget to bump script version as well.

flashcode:
Closing old PR without activity.
If needed, please open a new one and fill the new PR template accordingly.

PR #443, opened 2020-11-29
colorize_nicks.py 28: fix ignore_tags, broken by weechat 2.9 changes

jesopo:
simple fix: referencing `modifier_data` is no longer correct because it needs version-specific parsing, which was already being done.

I notice there's a couple of other `colorize_nicks.py` PRs - I'm happy to fix this PR if those are merged first!

### DIFF
--- a/python/colorize_nicks.py
+++ b/python/colorize_nicks.py
@@ -21,6 +21,8 @@
 #
 #
 # History:
+# 2020-11-29: jess
+#   version 28: fix ignore_tags having been broken by weechat 2.9 changes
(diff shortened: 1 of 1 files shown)


weechatter:
you don't mean this https://weechat.org/files/releasenotes/ReleaseNotes-devel.html#v2.9_modifier_weechat_print?
its already fixed with v27 (dd compatibility with new weechat_print modifier data)

jesopo:
the change in colorize_nicks.py that was done to support that modifier change broke the script

jesopo:
well, i guess more accurately, v27 didn't change everything that needed changing for the weechat 2.9 modifier change

PR #445, opened 2020-12-27
grep.py: ignore utf-8 decoding errors

anonymous2ch:
They prevent grep from working on logs, which contain incorrectly split international messages (i.e. message containing 242 utf-8 chars and 3 spaces (484 + 3 bytes) was split like 243+244 bytes, rendering utf-8 undecodable in the second message)


flashcode:
Hi,
Could you please also bump version in script and update the ChangeLog?
Thanks.

flashcode:
Merged with bump of version (0.8.6) and update of ChangeLog.

PR #446, opened 2020-12-27
url_olde.py: Python 3 fix

anonymous2ch:
urldefrag is no longer in urlparse, but rather in [urllib.parse](https://docs.python.org/3.7/library/urllib.parse.html)


flashcode:
Hi,
Could you please bump script version?
Thanks.

flashcode:
Merged, with version bumped to 0.7.

PR #448, opened 2021-01-20
Create auto_translate.py

anonymous2ch:
A script that automatically translates every incoming message on an IRC channel to target language via google translate

### DIFF
--- a/python/auto_translate.py
+++ b/python/auto_translate.py
@@ -1,0 +1,173 @@
+# -*- coding: utf-8 -*-
+###
+#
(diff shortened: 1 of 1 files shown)


flashcode:
Hi,
Closing this MR as the process to submit a script is the form on weechat.org, and this script is already pending for approval: https://weechat.org/scripts/pending/

anonymous2ch:
this was actually the same script, but before some kind man on #weechat irc channel told me the pending was not completely compliant with requirements... So I rewrote it (see commit history)

PR #452, opened 2021-02-24
slack.py 2.7.0: Merge with mainline

trygveaa:
This is version 2.7.0 of slack.py (currently the most recent), copied
over from the wee-slack repo.


trygveaa:
Slack will remove some APIs at the end of today, which means that older versions of wee-slack will stop working. Therefore, I would be grateful if this could be merged as quickly as possible.

PR #458, opened 2021-04-05
gnome_screensaver_away.py 0.2.1: Add missing call to infolist_free

grdryn:
## Script info

- Script name: gnome_screensaver_away.py
- Version: 0.2.1

- Requirements: dbus-python

- Min WeeChat version: 0.3.0

- Script tags: 

## Checklist (script update)

@flashcode I haven't marked this as Python 3 compatible yet, as I'm not 100% sure. I tested before and after this change and found a bug (unrelated to this change) where this script being loaded seems to prevent clean `/quit` of weechat. Since I've _only_ got Python 3 at the moment, that's where I've experienced it. I don't know whether it also currently happens with Python 2. I'll try to find some more time to figure it out, but it might take a while to get all the weechat scripting stuff back into my head again!

Since the issue I describe also happens with the existing version, I think it's still reasonable to consider merging this change.

### DIFF
--- a/python/gnome_screensaver_away.py
+++ b/python/gnome_screensaver_away.py
@@ -25,6 +25,9 @@
 #             - Allow configuration of message and poll time
 #             - Tolerate gnome-shell crashes
 #
+# 2021-04-05: Sébastien Helleu <flashcode@flashtux.org>
+#     0.2.1 : - Add missing call to infolist_free
+#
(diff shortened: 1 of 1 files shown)


flashcode:
The script itself is supposed to be Python 3 compliant, I'll add the missing tag `py3`.

PR #477, opened 2021-09-09
Rename queue.py to cmdqueue.py

jspricke:
Python 3 provides a queue.py so the name is reserved.

If an other scripts does an `import queue` weechat will log:
python: script "queue" already registered (register ignored)
python: unable to register script "queue" (another script already exists with this name)

Ref.: https://bugs.launchpad.net/ubuntu/+source/weechat-scripts/+bug/1903354

### DIFF
--- a/python/cmdqueue.py
+++ b/python/cmdqueue.py
@@ -1,5 +1,5 @@
 # ==================================================== #
-# Script Name: queue.py	
+# Script Name: cmdqueue.py
 # Script Author: walk <KingPython@gmx.com>
 # Script Purpose: Command queing at its finest. Hopefully.
 #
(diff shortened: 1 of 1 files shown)


jspricke:
I've tried to contact the script author but the mail address does not exist anymore.

weechatter:
i would recommend to rename the script wqueue.py

flashcode:
Or a short name like cmdqueue.py ?

jspricke:
I've updated it to cmdqueue.py :).

weechatter:
perfect! :-)

flashcode:
Thanks, merged (with update of script version and ChangeLog).

PR #488, opened 2021-12-02
soju.py 0.1.5: adapt all username settings

khardix:
## Script info

- Script name: soju.py
- Version: 0.1.5

## Description

1. Correctly slices and dices usernames that contain `@client` suffix. `user@client` becomes `user/network@client` for the generated connections.
2. Applies the same handling to `sasl_username` parameter.

## Checklist (script update)

---
cc @emersion for review :)

### DIFF
--- a/python/soju.py
+++ b/python/soju.py
@@ -5,7 +5,7 @@
 
 import weechat
 
-weechat.register("soju", "soju", "0.1.4", "AGPL3", "soju bouncer integration", "", "")
+weechat.register("soju", "soju", "0.1.5", "AGPL3", "soju bouncer integration", "", "")
 
 bouncer_cap = "soju.im/bouncer-networks"
 caps_option = weechat.config_get("irc.server_default.capabilities")
@@ -87,9 +87,6 @@ def handle_bouncer_msg(data, signal, signal_data):
     net_name = "".join(ch if check_char(ch) else "_" for ch in net_name)
 
     addr = weechat.config_string(weechat.config_get("irc.server." + server_name + ".addresses"))
-    username = weechat.config_string(weechat.config_get("irc.server." + server_name + ".username"))
-    if "/" in username:
-        username = username.split("/")[0]
     add_server = [
         "/server",
         "add",
@@ -97,10 +94,17 @@ def handle_bouncer_msg(data, signal, signal_data):
         addr,
         "-temp",
         "-ssl",
-        "-username=" + username + "/" + net_name,
     ]
 
-    for k in ["password", "sasl_mechanism", "sasl_username", "sasl_password"]:
+    # User name settings need to be adapted for new networks
+    for k in ["username", "sasl_username"]:
+        v = weechat.config_string(weechat.config_get("irc.server." + server_name + "." + k))
+        if not v:
+            continue
+        username = v.split("/", maxsplit=1)[0] + "/" + net_name
+        add_server.append("-" + k + "=" + username)
+
+    for k in ["password", "sasl_mechanism", "sasl_password"]:
         v = weechat.config_string(weechat.config_get("irc.server." + server_name + "." + k))
         add_server.append("-" + k + "=" + v)
 


emersion:
> Correctly slices and dices usernames that contain @client suffix. user@client becomes user/network@client for the generated connections.

Hm, but `user@client/network` should work just fine too. Or am I missing something?

khardix:
> > Correctly slices and dices usernames that contain @client suffix. user@client becomes user/network@client for the generated connections.
> 
> Hm, but `user@client/network` should work just fine too. Or am I missing something?

The [docs](https://man.sr.ht/chat.sr.ht/quickstart.md#multi-client-identification) state that is should be this way; did not occur to me to check the other way :sweat_smile: 

The root problem for me in the end was the application to `sasl_username` (or lack of thereof), since I did not have just `username` set at all.

emersion:
Okay, what about this?

- If `username` of the original server is empty, don't set it on the new server
- If `username` of the original server is not empty, strip `/<old network>` if any and append `/<new network>`
- Apply the same logic to `sasl_username`

khardix:
I will try to get back to this over the weekend.

khardix:
So, I managed to come back to this; from testing, it turned out that the `@client` suffix was not the issue (as correctly pointed out in [previous comment](https://github.com/weechat/scripts/pull/488#issuecomment-984502665)) – not adapting the `sasl_username` setting in addition to the `username` was. Hence the change in commit and PR name.

I re-did the patch focusing entirely on the bugfix in question, resisting the urge to refactor as I went along. Let me know if some code polishing would be also welcome (perhaps in a separate PR).

emersion (approved):
LGTM, thanks!

> Let me know if some code polishing would be also welcome (perhaps in a separate PR).

Sure! But with a disclaimer: I'm a bit opinionated, I like to keep things simple, and dislike "clever" code. I prefer readability and clarity over e.g. using fancy Python features only Python experts know about, or unnecessarily abstracting the code.

Additional note: it would be very nice to add `bouncer-networks-notify` support to this script, to add/remove networks on-the-fly. :P

emersion:
Oh, forgot one thing: can you bump the script version in the `weechat.register` call?

khardix:
> > Let me know if some code polishing would be also welcome (perhaps in a separate PR).
> Sure! But with a disclaimer: I'm a bit opinionated, I like to keep things simple, and dislike "clever" code. I prefer readability and clarity over e.g. using fancy Python features only Python experts know about, or unnecessarily abstracting the code.

I also tend to prefer readability,
but being (if I say so myself 😛) a bit of a Python expert,
my opinion on what's readable my differ from yours 😉.

However, if you are willing to review it and vet it,
I'm willing to take a stab at it.

> Additional note: it would be very nice to add `bouncer-networks-notify` support to this script, to add/remove networks on-the-fly. :P

No promises there, but I may take a look 🙂
Did not implement anything like this before, so… any RFCs to look at?

khardix:
> Oh, forgot one thing: can you bump the script version in the `weechat.register` call?

And that's what I missed when I re-did the patch :) Incoming.

emersion:
> However, if you are willing to review it and vet it, I'm willing to take a stab at it.

Nice, definitely!

> Did not implement anything like this before, so… any RFCs to look at?

soju's extension is documented at https://git.sr.ht/~emersion/soju/tree/master/item/doc/ext/bouncer-networks.md?view-source

The gist of it is:

- Enable the `soju.im/bouncer-networks-notify` (we onlly enable the `soju.im/bouncer-networks` capability right now). This will make it so the bouncer will send some IRC messages when a network is added, removed, or edited.
- Remove the `LISTNETWORK` command we send. The bouncer will send the current list of networks when connecting.
- Listen to these messages and update the WeeChat config as needed.
  - Network add should already be handled with the existing code.
  - We don't need to do anything on network change, but there's a twist: when a network's name changes, we need to avoid getting confused. We could rename the network, WeeChat seems to have a command to do that.
  - We need to remove the server when a network is removed. That's indicated with a `BOUNCER NETWORK <id> *` message.

PR #491, opened 2022-01-29
New script spoiler.py: send spoiler text

tfaughnan:
## Script info

- Script name: spoiler
- Version: 0.1.0

- Requirements: none

- Min WeeChat version: 0.3.7

- Script tags: irc, color

## Description

This script converts XML-inspired tags of the form `<spoiler>foo bar</spoiler>` in outgoing IRC messages into spoiler text by coloring the text between the tags with the same foreground and background color, using the widely-supported U+0003 IRC coloring scheme.

## Checklist (new script)

### DIFF
--- a/python/spoiler.py
+++ b/python/spoiler.py
@@ -1,0 +1,76 @@
+# Copyright (c) 2022 Thomas Faughnan <tom@tjf.sh>
+# Released under GNU GPLv3
+# https://www.gnu.org/licenses/gpl-3.0.html
+
+# =============================================================================
+#                               SPOILER.PY
+#
+# Send spoiler text (same foreground and background color) like so:
+#   The answer to life, the universe, and everything is <spoiler>42</spoiler>.
+# An optional color integer in [0, 98] or color name may be provided like so:
+#   <spoiler 3>Darth Vader</spoiler> is <spoiler blue>Luke</spoiler>'s father.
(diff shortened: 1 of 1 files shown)


python/spoiler.py
```diff
@@ -0,0 +1,76 @@
...
+    prefix, body = m.groups()
+
+    return prefix + re.sub(SPOILER_PATTERN, repl, body)
+
+
+if __name__ == '__main__' and import_ok:
+    weechat.register('spoiler', 'Thomas Faughnan', '0.1.0', 'GPL3',
+                     'Send spoiler text <spoiler>like this</spoiler>', '', '')
+    weechat.hook_modifier('input_text_for_buffer', 'spoilerize_buffer', '')
+    weechat.hook_modifier('irc_out_privmsg', 'spoilerize_irc', '')
```
flashcode, line 76: Hi,
If I'm not wrong, this modifier specific to IRC is not needed there: the "input_text_for_buffer" works on all buffers, including IRC channels/privates.
If confirmed, could you please remove it (as well as its callback)?
Thanks.

flashcode (changes requested):

python/spoiler.py
tfaughnan: I actually don't really have a use case for this script anymore, so I'm going to close my PR.

PR #505, opened 2022-09-05
go.py: use buf number instead of name

craftyguy:
## Script info

- Script name: go.py 
- Version: 

- Requirements: 

- Min WeeChat version: 

- Script tags: 

## Description

This fixes some issues when switching to a buffer, when the buffer's
name is something that weechat does not like, e.g. "#Foo < Bar"

    weechat =!= Error with command "/buffer #Foo < Bar" (help on command: /help buffer)

These aren't really valid channel names(I think?), but some weechat
plugins (weechat-matrix) may create buffers with incompatible names like
that. In any case, switching based on the buffer *number* seems to be a
reliable compromise.

### DIFF
--- a/python/go.py
+++ b/python/go.py
@@ -294,11 +294,11 @@ def go_now(buf, args):
         for index in range(len(listbuf)):
             if go_match_beginning(listbuf[index], args):
                 weechat.command(buf,
-                                '/buffer ' + str(listbuf[index]['full_name']))
+                                '/buffer ' + str(listbuf[index]['number']))
                 return
 
     # jump to first buffer in matching buffers by default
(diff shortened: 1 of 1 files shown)


flashcode:
Hi,

Thanks for the PR.

Switching to the number has not the same effect as the name: if buffers are merged, using the number may not activate the target buffer.

So I would not change that, but instead fix plugins/scripts that open buffers with bad name, or maybe better, make WeeChat refuse completely to create these buffers, which will force plugins/scripts to sanitize names before creating buffers.

flashcode:
By the way your command example `/buffer #Foo < Bar` works for me (I created such test buffer), no such error.
Which version of WeeChat are you using?

flashcode:
Changes are not needed for me, so I close the pull request.

PR #518, opened 2022-12-15
urlgrab.py: Use xdg-open as default local command

3v1n0:
## Script info

- Script name: `urlgrab.py`
- Version: 3.2

## Description

It will automatically open the URI in the default browser for the system, without having users to change it to their browser.

## Checklist (script update)


flashcode:
Thanks!

PR #519, opened 2023-01-17
New script digraph.py: nvim style digraphs. Enables entering math sym…

narodnik:
…bols easily.

## Script info

- Script name:  digraph
- Version: 0.1

- Requirements: None

- Min WeeChat version: N/A

- Script tags: None

## Description

This script enables users to easily enter unicode digraphs for math symbols. When they type $, then everything after that will be replaced according to the digraphs until the next $ sign.

See here for a full list of them:

https://github.com/narodnik/weechat-digraph-latex

Will be updated over time + will eventually add a way to customize them when I get time.

## Checklist (new script)

### DIFF
--- a/python/digraph.py
+++ b/python/digraph.py
@@ -1,0 +1,172 @@
+# coding=utf-8
+# WeeChat script for nvim style digraphs. Enables entering math symbols easily.
+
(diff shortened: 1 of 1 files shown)


flashcode:
Thanks, merged!

PR #545, opened 2023-10-16
fish 0.15: fix encoding errors with message markers

freshprince:
## Script info

- Script name: fish.py
- Version: 0.15

## Description

- fix encoding errors with message markers (provided by @andkem)

## Checklist (script update)

- [X] Author has been contacted
- [X] Single commit, single file added
- [X] Commit message format: `script_name.py X.Y: …`
- [X] Script version and Changelog have been updated
- [X] For Python script: works with Python 3 (Python 2 support is optional)

### DIFF
--- a/python/fish.py
+++ b/python/fish.py
@@ -61,7 +61,7 @@ from os import urandom
 
 SCRIPT_NAME = "fish"
 SCRIPT_AUTHOR = "David Flatz <david@upcs.at>"
-SCRIPT_VERSION = "0.14"
+SCRIPT_VERSION = "0.15"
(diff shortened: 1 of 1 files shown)


trygveaa:
Which version of Python are you using in WeeChat? Run `/python version` to check.

I haven't tested this, but from looking at the code I don't see a reason to do this change with Python 3.

trygveaa:
Nevermind, I see that this is just a change from the upstream repo and that there's [an explanation](https://github.com/freshprince/weechat-fish/pull/28) there. (I also see that the encode/decode stuff isn't because of Python 2, but because you're dealing with binary data).

PR #546, opened 2023-10-27
New script greentext.py: colorize imageboard-style text formatting

agvxov:
## Script info

- Script name: greentext
- Version: v1

- Script tags: color, display, py3

## Description
Colorize imageboard-style text formatting.
Both inbound and outbound messages are colored.
Since the coloring uses IRC color codes,
outbound greentexting will be visible to both you and your friends.

The following are supported:
+ greentext
+ purpletext
+ redtext

## Checklist (new script)

### DIFF
--- a/python/greentext.py
+++ b/python/greentext.py
@@ -1,0 +1,78 @@
+# # Weechat-Greentext
+# Version:          v1
+# Min WeeChat
+#  version tested:  3.8
+# Author:           AGVXOV
+# Contact:          agvxov@gmail.com
+# Project Home:     https://github.com/agvxov/weechat-greentext
+# This script is Public Domain.
+#
+# Weechat script for applying imageboard formatting to messages. The following are supported:
+#  + greentext
+#  + purpletext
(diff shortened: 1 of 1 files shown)


python/greentext.py
```diff
@@ -0,0 +1,78 @@
...
+		else:
+			s = weechat.color("green") + s
+	return s
+
+def hi_purpletext(modifier, s):
+	if purpletext_re.search(s):
+		if modifier == 'irc_out1_PRIVMSG':
+			s = COLOR_PURPLE + s
+		else:
+			weechat.color("purple") + s
```
flashcode, line 50: `purple` is not a valid WeeChat color (see: https://weechat.org/files/doc/weechat/stable/weechat_user.en.html#command_line_colors), you meant `magenta`?

flashcode (changes requested):

agvxov:
Yeah, that was silly of me.
Should be good now.

flashcode (approved):

PR #551, opened 2024-03-05
Adds auto_read_buffer which marks a buffer as read upon switching

kindrowboat:
## Script info

- Script name: auto_read_buffer.py
- Version: 1.0.0

- Requirements: 

- Min WeeChat version: 

- Script tags: buffer, read

## Description

Automatically marks a buffer as read upon switching away from it to another buffer.

## Checklist (new script)

## Checklist (script update)

- [ ] Single commit, single file added
- [ ] Commit message format: `script_name.py X.Y: …`
- [ ] Script version and Changelog have been updated

## Checklist (script deletion)

- [ ] Single commit, single file deleted
- [ ] Commit message format: `Remove script name.py` with reasons in description

### DIFF
--- a/python/auto_read_buffer.py
+++ b/python/auto_read_buffer.py
@@ -1,0 +1,33 @@
+# -*- coding: utf-8 -*-
+import weechat
+
+SCRIPT_NAME = "auto_read_buffer"
+SCRIPT_AUTHOR = "Stef Dunlap <hello@kindrobot.ca>"
+SCRIPT_VERSION = "1.0"
+SCRIPT_LICENSE = "GPL3"
+SCRIPT_DESC = "Automatically marks a buffer as read upon switching away"
+
+# Global variable to store the current buffer
+current_buffer = None
+
+# Callback for the buffer switch signal
+def buffer_switch_cb(data, signal, signal_data):
+    global current_buffer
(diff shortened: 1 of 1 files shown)


flashcode:
Hi,
Sorry for the delay, does this script fix a bug in WeeChat?
What's the use case where the buffer is not marked as read?

kindrowboat:
Hey, no problem! I think this is an issue when you're using a weechat relay. Before this script, when I moved off of a buffer, it was still marked yellow/unread. I'd have to switch  off and back onto to the buffer to get it to to be marked as read. This is true for only the relay host, I think. I swear I saw a known issue/wontfix for this, but I can't seem to find it now.

flashcode:
OK, for relay (weechat protocol), this is caused by the option `weechat.look.hotlist_add_conditions` whose default value is `${away} || ${buffer.num_displayed} == 0 || ${info:relay_client_count,weechat,connected} > 0`.
So the last condition means if there's at least one client of the weechat relay, always add the buffer to the hotlist, even if it's the current one.
It's done like this because the relay client may not see the same buffer as active, and thus you don't miss any notification in the client.
If needed, you can remove this condition, but then that means the weechat current buffer will not be added to hotlist and not visible in the client.

PR #554, opened 2024-05-04
buffer_autoclose.py 0.6: add setting to prefer closing some buffers

vmiklos:
## Script info

- Script name: buffer_autoclose.py
- Version: 0.6

## Description

E.g. the Matrix script maps private rooms with multiple people to buffers in weechat which are nominally not private. This means that these rooms pile up, which is normally solved for real 1:1 private chats.

Fix the problem by adding a way to autoclose also non-private buffers.

No behavior is changed unless the new config option is set.

## Checklist (script update)

### DIFF
--- a/python/buffer_autoclose.py
+++ b/python/buffer_autoclose.py
@@ -20,6 +20,8 @@
 # (this script requires WeeChat 0.3.0 or newer)
 #
 # History:
+# 2024-05-04, Miklos Vajna
+#  version 0.6: Allow autoclosing explicitly listed non-private buffers as well
 # 2018-04-10, Sébastien Helleu <flashcode@flashtux.org>
 #  version 0.5: fix infolist_time for WeeChat >= 2.2 (WeeChat returns a long
 #               integer instead of a string)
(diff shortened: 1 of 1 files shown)


python/buffer_autoclose.py
```diff
@@ -66,6 +69,10 @@ def get_all_buffers():
     infolist = w.infolist_get('buffer', '', '')
     while w.infolist_next(infolist):
         buffer_type = w.buffer_get_string(w.infolist_pointer(infolist, 'pointer'), 'localvar_type')
+        name = w.buffer_get_string(w.infolist_pointer(infolist, 'pointer'), 'name')
+        if name in w.config_get_plugin('prefer').split(','):
```
flashcode, line 73: For optimization reason, you should perform this split before the while loop, so it's done only once (as the value of option does not change during this loop).

flashcode (changes requested):

python/buffer_autoclose.py
vmiklos: Makes sense, thanks for noticing. :-)

vmiklos:
Hi, anything else I should do to get this merged? Thanks.

flashcode (approved):

PR #571, opened 2025-03-01
New script auto_commands.py: Auto sends commands on client start

tomteipl:
## Script info

- Script name:  `auto_commands`
- Version: `0.3`

- Script tags: `py2` `py3` `config` 

## Description

Create and add your own **commands** to be send on client start. 
You can specify after what time the commands will be send, for example: after 10 sec.
Check list of added commands by you and remove them by **index** or **text**.

## Checklist (new script)

- [ ] Function `hook_process` is used for any blocking call

### DIFF
--- a/python/auto_commands.py
+++ b/python/auto_commands.py
@@ -1,0 +1,165 @@
+# Copyright (c) 2025 by Kamil Wiśniewski <tomteipl@gmail.com>
+#
+# This script sends auto commands on start
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+#
+#[Change Log]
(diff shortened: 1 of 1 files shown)


weechatter:
Have a look at  weechat.startup.command_after_plugins and weechat.startup.command_before_plugins
Especially weechat.startup.command_before_plugins do not need any plugin (eg python).
Sorry to say, but your script seems obsolete for me.

tomteipl:
> Have a look at weechat.startup.command_after_plugins and weechat.startup.command_before_plugins Especially weechat.startup.command_before_plugins do not need any plugin (eg python). Sorry to say, but your script seems obsolete for me.

You are right. I should gather more information about `weechat` as im fairly new with it.
Also im learning and practicing so every advise and review is helpful.

weechatter:
If you have an idea, simply ask in libera.#weechat if there is alrady an solution ;-)

flashcode:
I agree with @weechatter, the default options `weechat.startup.command_before_plugins` and `weechat.startup.command_before_plugins` are made to run commands on startup.

So I think your script is not needed, or maybe I missed something?

tomteipl:
@flashcode 
Yup its not necessary. I didnt know there were options like these for **Weechat**.

Although its easier at least for me to have simple separate list of commands i added. 
It also has potential for more features like suspending commands from the list or other things i have in mind.

I will keep it in my public repo and continue developing. 

I appreciate the answer, **PR can be closed sir.**

PR #572, opened 2025-03-01
New script ollama.py: AI auto-responder using Ollama

prietus:
This script automatically responds to mentions and private messages in WeeChat using Ollama LLM.
It also allows manual queries with the command `/ollama <question>`.

- Supports configuring auto-responses (`/set plugins.var.python.ollama_bot.highlight_response on/off`).
- Uses the WeeChat Python API and makes requests to a locally running Ollama instance.


flashcode:
Hi,

Thank you for the script.

Before your script can be integrated, please:

- squash all your commits in a single one
- include in your PR description the checklist (it is displayed when you make a new PR)
- please use [hook_url](https://weechat.org/files/doc/devel/weechat_plugin_api.en.html#_hook_url) (WeeChat ≥ 4.1.0) or [hook_process_hashtable](https://weechat.org/files/doc/devel/weechat_plugin_api.en.html#_hook_process_hashtable) to fetch URL (otherwise it would block WeeChat)

e -i HEAD~3
```diff
@@ -0,0 +1,2134 @@
+[33m0e021b6[m[33m ([m[1;36mHEAD[m[33m -> [m[1;32madd-ollama-bot[m[33m)[m New script ollama_bot.py: Adds non-blocking API calls using hook_process_hashtable()
```
flashcode, line 1: It seems this file has been committed by error, please remove it.

python/ollama.py
```diff
@@ -0,0 +1,146 @@
+import weechat
```
flashcode, line 1: Please add SPDX tags for copyright/license, which have been made mandatory in the meanwhile (see [Contributing guide](https://github.com/weechat/scripts/blob/main/CONTRIBUTING.md#copyright-and-license)), for example:

```
# SPDX-FileCopyrightText: 2025 teraflops <cprieto.ortiz@gmail.com>                                                                                                                            
#                                                                                                                                                                                      
# SPDX-License-Identifier: MIT
```

python/ollama.py
```diff
@@ -0,0 +1,146 @@
...
+Ollama Bot for WeeChat (Non-blocking version)
+
+This script automatically responds to mentions in channels and private messages using an Ollama LLM running locally.
+
+Features:
+- Responds to mentions in channels.
+- Can respond to private messages if enabled.
+- Allows manual queries using the /ollama command.
+- Configurable via WeeChat /set commands.
+- Uses hook_url for non-blocking HTTP requests.
```
flashcode, line 14: This comment says it uses `hook_url`, but `hook_process_hashtable` is used instead.

Using hook_url has 2 advantages:

- it does not rely on an external curl command (which may not be available), but it still uses curl internally (linked with weechat)
- it uses a thread instead of a process, which is more lightweight.

flashcode (changes requested):

PR #573, opened 2025-03-26
New script colorize_regex.pl: colorize highlight regex matches in chat messages

ryoskzypu:
## Script info

- Script name: colorize_regex.pl
- Version: 1.0

- Requirements: 

- Min WeeChat version: 

- Script tags: color, highlight, irc, regex

## Description

Colorize regex matches in chat messages from 'weechat.look.highlight_regex' option or the 'highlight_regex' buffer property, to make validation of matches easier.

## Checklist (new script)

- [ ] Function `hook_url`, `hook_process` or `hook_process_hashtable` is used for any blocking call

### DIFF
--- a/colorize_regex.pl
+++ b/colorize_regex.pl
@@ -1,0 +1,1378 @@
+# SPDX-FileCopyrightText: 2025 ryoskzypu <ryoskzypu@proton.me>
+#
+# SPDX-License-Identifier: MIT-0
+#
+# colorize_regex.pl — colorize highlight regex matches in chat messages
+#
+# Description:
+#   Colorize regex matches in chat messages from 'weechat.look.highlight_regex'
+#   option or the 'highlight_regex' buffer property, to make validation of matches
+#   easier.
+#
+#   Commands:
+#     /dbgcolor
+#       Debug weechat's strings color codes. Similar to '/debug unicode', but
+#       specific to colors, and output is a standard hex dump. Can be used to
+#       debug other scripts.
+#       See '/help dbgcolor' for usage.
+#
+#     /rset
+#       Fast set the script options and regexes. Similar to /fset.
+#       See '/help rset' for usage.
+#
+# Bugs:
+#   https://github.com/ryoskzypu/weechat_scripts
+#
+# History:
+#   2025-03-26, ryoskzypu <ryoskzypu@proton.me>:
+#     version 1.0: initial release
+
+use strict;
+use warnings;
+
+# Debug data structures.
+#use Data::Dumper qw< Dumper >;
+#$Data::Dumper::Terse = 1;
+#$Data::Dumper::Useqq = 1;
+
+# Global variables
+
+my %script = (
+    prog    => 'colorize_regex',
+    version => '1.0',
+    author  => 'ryoskzypu <ryoskzypu@proton.me>',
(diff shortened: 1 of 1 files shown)


weechatter:
how about improving colorize_lines.pl?
colorize_lines.pl  can do "highlight word(s) in text, matching word(s) in weechat.look.highlight" with option , /fset plugins.var.perl.colorize_lines.highlight_words

weechatter:
maybe its better to add the regex function to colorize_lines.pl, to have one script with nearly same functionality.  what do you think @flashcode ?

flashcode:
It depends if it's simple to add into colorize_lines.pl, and it could prevent duplicated code.
But on the other hand, if it brings many other options, a separate script could make sense.

ryoskzypu:
> It depends if it's simple to add into colorize_lines.pl, and it could prevent duplicated code. But on the other hand, if it brings many other options, a separate script could make sense.

The scripts are different in many ways, e.g. colorize_regex has a command like /debug unicode, a dedicated buffer that shows commands/debug output, uses hook_line rather than hook_modifier, uses config_new(), etc.

I could add all the features to colorize_lines.pl, but I don't know if it's okay; or only the highlight_regex function.

flashcode:
@ryoskzypu: then let's keep the separate script.

perl/colorize_regex.pl
```diff
@@ -0,0 +1,1396 @@
+# SPDX-FileCopyrightText: 2025 ryoskzypu <ryoskzypu@proton.me>
+#
+# SPDX-License-Identifier: MIT-0
+#
+# License:
+#   MIT No Attribution
+#
+#   Copyright (c) 2025 ryoskzypu <ryoskzypu@proton.me>
```
flashcode, line 8: This copyright line should be removed (duplicate with the SPDX copyright tag above) as it would generate a duplicate copyright line in the SPDX bill of materials:

```
$ reuse spdx

(...)

FileName: ./perl/colorize_regex.pl
SPDXID: SPDXRef-f61225f4475537305dbb661a5ed0d658
FileChecksum: SHA1: cc0594b6b3513c733fbc41316dc0007716ba94dd
LicenseConcluded: NOASSERTION
LicenseInfoInFile: MIT-0
FileCopyrightText: <text>Copyright (c) 2025 ryoskzypu <ryoskzypu@proton.me>
SPDX-FileCopyrightText: 2025 ryoskzypu <ryoskzypu@proton.me></text>

(...)
```

perl/colorize_regex.pl
```diff
@@ -0,0 +1,1396 @@
...
+#
+#   Permission is hereby granted, free of charge, to any person obtaining a copy of this
```
flashcode, line 10: The complete license text will be available in the repository (in LICENSES/ directory), so having the license text in the script itself is optional.

You can safely remove it, but you can also keep it if you absolutely want it there (not a problem).

flashcode (changes requested):

ryoskzypu:
@flashcode: Done, I think it is ok now? Sorry for the force-push noise, I've messed up when rebasing my fork...

flashcode:
Thanks, script merged (I moved script to perl/ directory).

PR #574, opened 2025-05-08
colorize_nicks.py 33: add many improvements, features, and fixes

### LINKED ISSUE #49: [colorize_nicks.py] IRC color overriding
This plugin doesn't work too well when applied to messages which contain IRC color codes. For example:
- nick isn't colored when IRC color code is right in front of it or in the middle of it
- when nick is in a part of the message with IRC color set and is colorized, the color after the nick is "reset" not the one which was supposed to be there

### LINKED ISSUE #70: colorize_nicks.py prefixes all prefixless nicks with F10
- ircs://irc.chatspike.net:6697/#Falling
- https://github.com/weechat/weechat/issues/427

screenshot — http://i.imgur.com/zXlGzGJ.png

### LINKED ISSUE #133: [colorize_nicks.py] Freeze on lines with repeated nick and greedy_matching = on
[colorize_nicks.py](https://github.com/weechat/scripts/blob/master/python/colorize_nicks.py) seems to hang with a CPU at 100% while matching a line with a nick repeated a lot of times.

An example of line causing the problem and the backtrace (normal and full) are available [here](https://gist.github.com/bendem/016a97540dc226730335).

### LINKED ISSUE #135: [colorize_nicks.py] Mentioning nick "F1" breaks line prefix
Being in a channel with someone with the nick "F1". If that nick gets mentioned, their prefix will be broken, i.e. "F10" prepended: 

It seems like the script is interfering with a color code in the line prefix.

Unloading colorize_nicks.py solves the issue. The problem can currently be avoided by adding "F1" to `colorize_nicks.look.blacklist_nicks`.

_Brought to my attention by lunchdump in #weechat._

### LINKED ISSUE #175: colorize_nicks breaks the surrounding colors
If the line already looks like this: [yellow]Hello nick nice to see you[/yellow]

And colorize_nicks wants to do this: [purple]nick[/purple]

You end up with this: [yellow]Hello [/yellow][purple]nick[/purple] nice to see you

Instead of just adding a generic color-reset character, it should look at the whole line and repeat whatever color control characters came before it. This is more complicated but more correct.

ryoskzypu:
## Script info

- Script name:  colorize_nicks.py
- Version:  33

- Requirements: 

- Min WeeChat version: 

- Script tags: 

#### Changes:
 - Remove the `VALID_NICK` regex to not be dependent on RFC spec and change the loop to split only on spaces, so the nick regex is free to match any suffix.
 - Add `nick_prefixes` and `nick_suffixes` options to configure charset affixes.

 - Replace the default 'greedy' matching with improved 'lazy' matching, because 'greedy' matches nicks incorrectly and crashes weechat. 
   - 'lazy' matching does not match nicks in substrings, so remove `ignore_nicks_in_urls` option.
   - fixes #133 and #197.

   - Rationale:
     - The greedy matching code recolors duplicated strings because of the way `replace()` works, so it crashes weechat on multiple duplicated nicks.
     - #259 does not really prevent that, since it can still be reproduced by mentioning a short nick e.g. F1, repeatedly in a message.
     - It can be fixed by iterating on unique words, but it still colorizes nicks in substrings if a nick is mentioned before (e.g. alice https://www.alice.com), hence the option `ignore_nicks_in_urls` fails.

 - Replace `hook_modifier()` with `hook_line()` for a more granular parsing (e.g. prefix is separated from message, get filtered lines, etc).
   - fixes #70, fixes #135.

 - Add preserving of message colors logic.
   - fixes #49, fixes #175.
 - Add `irc_only` option to ignore non IRC messages (i.e. set buffer restrictions: plugin = irc, tags = irc_privmsg and irc_notice, type = channel and private).
 - Add do not colorize nicks in filtered messages, and `colorize_filtered` option.
 - Add colorizing of IRC prefixes and nicks on IRC private buffers.
 - Update `colorized_input_cb()`  with the changes, and add decoding of IRC colors from input.
 - Update `config_init()` parsing of options.
 - Remove 'so,root' from `blacklist_nicks` default option. (These nicks do not make sense for defaults.)
 - Change `min_nick_length` default option to 1. (Single char nicks should not be excluded by default.)
 - Remove `hook_modifier()` that has the modifier colorize_nicks. (Is it possible to pass custom modifiers?)
 - Replace `%` formatting strings with f-strings, `"` with `'`, and some code to use the `:=` operator.
 - Update identifiers, comments, and add some whitespace in code for readability.
 - Remove unused global keywords and add error handlings.
 - Remove utf-8 encoding header, add SPDX copyright + license tags, and add upstream link.

## Checklist (script update)

### DIFF
--- a/python/colorize_nicks.py
+++ b/python/colorize_nicks.py
@@ -1,6 +1,7 @@
-# -*- coding: utf-8 -*-
+# SPDX-FileCopyrightText: 2010 xt <xt@bash.no>
+# SPDX-FileCopyrightText: 2025 ryoskzypu <ryoskzypu@proton.me>
 #
-# Copyright (c) 2010 by xt <xt@bash.no>
+# SPDX-License-Identifier: GPL-3.0-or-later
 #
 # This program is free software; you can redistribute it and/or modify
 # it under the terms of the GNU General Public License as published by
@@ -20,7 +21,12 @@
 # not just in the prefix section.
 #
 #
+# Bugs:
+#   https://github.com/ryoskzypu/weechat_scripts
+#
 # History:
+# 2025-05-08: ryoskzypu <ryoskzypu@proton.me>
+#   version 33: add many improvements, features, and fixes
 # 2023-10-30: Sébastien Helleu <flashcode@flashtux.org>
 #   version 32: revert to info "nick_color" with WeeChat >= 4.1.1
 # 2023-10-16: Sébastien Helleu <flashcode@flashtux.org>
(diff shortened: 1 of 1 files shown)


weechatter:
works for me

flashcode:
Hi @ryoskzypu,

Thanks for all these improvements.

To trace the history, could you please add all new features and fixes into the body of the 
commit message (using `(closes #1234)` to close bugs).

And then you can remove `(see Changes in the PR #574 for details)` fro the script comment, as it's better to have detail of changes in git itself.

Thanks!

ryoskzypu:
@weechatter: Thanks for testing.

@flashcode: Done! Is it okay? I tried my best to fit the changes in commit for git log.

> Thanks for all these improvements.

You're welcome and thank you for this great piece of software ;-).

flashcode:
Thanks, merged.

serious-angel:
Thank you, heartfelt, for the awesome work, marvel you do, dear Developers and Artists! :sparkles:

PR #591, opened 2026-01-10
smarter_tts.py

brhacket:
## Script info

- Script name: `smarter_tts.py`
- Version: `1.6`

- Requirements: `speech-dispatcher` (requires `spd-say` binary in PATH)

- Min WeeChat version:

- Script tags: `tts`, `sound`, `audio`, `accessibility`, `speech`

## Description

A zero-latency, thread-safe Text-to-Speech integration using the system's `speech-dispatcher`.

**Why this is needed (vs existing `tts.py`):**
This script is a modern architectural alternative to the existing `tts.py`.
1. **Threading vs Process Hook:** Unlike `tts.py` which spawns a new process for every message (causing UI freezes on high load), this script uses a persistent background thread and a FIFO queue.
2. **Standardization:** It targets the Linux standard `speech-dispatcher` rather than hardcoding specific binary flags for `espeak` or `festival`.
3. **Smart Logic:** It includes "Burst Merging" (combining rapid messages from the same user) and "Nick Humanization" (cleaning up nicknames for phonetic pronunciation) which are not present in older scripts.

**Key Features:**
- **Auto-Discovery:** Automatically scans the system to find sound files for notifications on any distro (Fedora, Arch, Debian, etc).
- **Focus Mode:** Only reads the active buffer by default, but allows locking to specific channels via `/listento`.
- **Background Alerts:** Plays a sound and announces channel names for background highlights without reading private text.

## Checklist (new script)

- [x] Commit message: `New script smarter_tts.py: Zero-latency TTS integration`
- [x] Comment in script with name/pseudo, e-mail and license using [SPDX](https://spdx.dev/) tags (see [Contributing guide](https://github.com/weechat/scripts/blob/main/CONTRIBUTING.md#copyright-and-license))
- [x] Function `hook_url`, `hook_process` or `hook_process_hashtable` is used for any blocking call
    *   *(Note: Script uses standard python threading for the audio queue to ensure zero-latency continuity without blocking the main loop, utilizing `hook_timer` to safely update UI from the main thread).*


python/tts_weechat_by_brhacket.py
```diff
@@ -0,0 +1,415 @@
+# -*- coding: utf-8 -*-
```
flashcode, line 1: The coding is not needed any more with Python 3.

Could you please rename the file to smarter_tts.py?

python/tts_weechat_by_brhacket.py
```diff
@@ -0,0 +1,415 @@
...
+#
+# smarter_tts.py - Zero-latency TTS with Auto-Discovery Sounds
+#
+# Copyright (c) 2025 brhacket <https://github.com/brhacket>
```
flashcode, line 5: Please convert to SPDX copyright and include your email for the contact.

The GitHub link may follow on another line (outside SPDX tags).

python/tts_weechat_by_brhacket.py
```diff
@@ -0,0 +1,415 @@
...
+import weechat
+import subprocess
+import threading
+import queue
+import time
+import os
+import shutil
+import re
+
+weechat.register("smarter_tts", "brhacket", "1.6", "GPL3", "Zero-latency TTS with Auto-Discovery", "", "")
```
flashcode, line 38: First version submitted could be 1.0 (except if you want to follow the version of an upstream script).

flashcode (changes requested):

PR #595, opened 2026-01-28
fish.py 1.0

freshprince:
## Script info

- Script name: fish.py
- Version: 1.0

- Requirements: pycryptodome

- Script tags: irc, crypto, fish py3

## Description

Handle keys with cbc: prefix and implement encryption in cbc mode 
Set server so that we send our key exchange reply correctly 
Handle DH1080 exchange in cbc-mode
Use string_eval_expression when deriving key so that we can use ${sec.data.foo} for keys 
Derive keys from weechat settings
Annouce whether stuff is encrypted in ecb or cbc mode 
Provide bar item that shows current buffer mode (ecb, cbc, plaintext) Change some default values for bar item color
Add different option for bar item indicator
Make sure we create the new buffer on the correct server 
Show different color in indicator when we don't know the current state 
Store encryption state in buffer localvar
Announce plaintext when state was unknown before
Add note about secure data
Use weechat to parse irc lines
Provide a command to add the bar item to weechat.bar.status.items 
Add information about fish in irc tags and use a line hook to show it in the prefix Don't encrypt empty messages
Make sure we remove fish tag of plaintext messages so that server cannot fool us 
Disable announce by default since prefix marker are nicer Try to import Cryptodome first
Add second DH1080_FINISH step to key exchange that other implementations seem to do
Try better to decrypt CBC messages that were cut off

See

  https://github.com/freshprince/weechat-fish/compare/v0.15-weechat-scripts...v1.0-weechat-scripts

for full changelog

## Checklist (script update)

We're using `irc_in_` functions because we want decrypt messages before weechat determines the encoding.

### DIFF
--- a/python/fish.py
+++ b/python/fish.py
@@ -1,6 +1,6 @@
 # -*- coding: utf-8 -*-
 #
-# Copyright (C) 2011-2022 David Flatz <david@upcs.at>
+# Copyright (C) David Flatz <david@upcs.at>
 # Copyright (C) 2017-2020 Marcin Kurczewski <rr-@sakuya.pl>
 # Copyright (C) 2017 Ricardo Ferreira <ricardo.sff@goatse.cx>
 # Copyright (C) 2014 Charles Franklin <jakhead@gmail.com>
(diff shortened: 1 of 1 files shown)


flashcode:
Thanks, merged.

I updated the copyright lines to use SPDX tags.

PR #602, opened 2026-06-08
fix: use raw string for glitter regex pattern

spithash:
Replace the regex pattern string with a raw string and remove unnecessary escaping of '*' in the character class.

This eliminates the warning emitted by newer python versions:

SyntaxWarning: "\*" is an invalid escape sequence. Such sequences will not work in the future.
Did you mean "\\*"?

The regex behavior is unchanged.


spithash:
I've looked at the author's profile and there's no public repo for the glitter.py plugin.

flashcode:
Thanks, merged.

PR #614, opened 2026-09-12
New script tennis.py: live tennis scores and ATP/WTA rankings on demand

bensynapse:
**Disclosure: I work on the Live Tennis API, the service this script talks to.** I am submitting it because the catalogue has no tennis or live-scores script of any kind, but please weigh it knowing the author has a commercial interest in the API. It needs a key from that service to do anything; free keys require no card. Happy to adjust or withdraw if a vendor-authored script is not something you want to carry. AI tooling was used while writing it; I understand every line and have tested the logic, as described below.

## Script info

- Script name: tennis.py
- Version: 1.0.0
- Requirements: an API key from https://livetennisapi.com (free tier available); no Python packages beyond the standard library
- Min WeeChat version: 4.1.0 (for `hook_url`)
- Script tags: tennis, sport, scores, api

## Description

Live tennis scores and rankings on demand, printed on the buffer where you ran the command.

- `/tennis live` — every match in play, one line each: match id, surnames, set scores, the points of the game in progress, then the tournament. `*` follows the player serving; `BP` means the receiver is one point from a break.
- `/tennis match <id>` — one match, with both players' country and ranking and the match state.
- `/tennis atp [count]` / `/tennis wta [count]` — singles rankings with the weekly movement.

Example line:

```
187701  Djokovic* v Alcaraz  6-4 3-3  30-40  BP  Wimbledon, Round of 16
```

Notes on the design:

- **Nothing on a timer.** The free tier allows 30 requests a minute and 100 a day, so the script only touches the network when you type a command. It installs no timer and polls nothing.
- **Non-blocking**, via `hook_url` with a callback; the client is never held up.
- The key lives in `plugins.var.python.tennis.api_key` and is read through `string_eval_expression` at call time, so `/secure` works: `/set plugins.var.python.tennis.api_key "${sec.data.livetennisapi}"`.
- Output goes back to the buffer the command was run on, resolved by full name at callback time rather than by holding a buffer pointer across the request.
- Break point is derived, not fetched: receiver at advantage, or receiver at 40 while the server is at 0/15/30; never inside a tiebreak. The score fields are nullable on a match that has just finished, and every formatter is null-safe.
- With no key set, one line names the option to set and says a free key exists. Non-2xx, unreadable bodies, rate limiting and plan limits each print a single readable line.
- **Two of the four commands need a paid plan on that API.** `/tennis atp` and `/tennis wta` answer "this data is not included in your plan" on a free key. `/help tennis` says so. `live` and `match` are the free-tier commands.

## Checklist (new script)

- [x] Comment in script with name/pseudo, e-mail and license using SPDX tags
- [x] Function `hook_url` is used for any blocking call
- [x] For Python script: works with Python 3
- [x] Score 100 / 100 displayed by weechat-script-lint

```
$ weechat-script-lint python/tennis.py
python/tennis.py: score = 100 / 100
Perfect: 1 scripts analyzed, 0 with issues: 0 errors, 0 warnings, 0 info
```

`make check` and `reuse lint` are both green locally (381/381 files with licence information).

One caveat for your review: I do not have WeeChat installed on the machine I wrote this on, so while the logic is covered by a stubbed-API harness (break-point derivation, score formatting, response and error handling, request construction, the version guard), I have **not** been able to `/script load` it in a real client. I would be grateful if a reviewer exercised that path.


flashcode:
Thanks, script added!

